PR #42, opened 2014-12-17
Use Pact-JVM for AccountClientTest

seanf:
Proof of concept for using Pact-JVM.


seanf:
Pact can't do what I had hoped (test a sequence of interactions which change the server state), so I don't think we'll worry about it.

PR #43, opened 2014-12-24
rhbz1124630 - need to add header for all requests

huangp:


huangp:
@aeng can you review this one please? Thanks

aeng:
:+1:

huangp:
@definite can you test this please? 
- [ ] zanata init works
- [ ] put account works
- [ ] copy trans works

huangp:
last commit fixed an overlooked bug:
when you select a project and a version using zanata init, if the project and version don't have project type specified, it will run into exception. Now this is fixed.
Possible test cases for zanata init:
- [ ] run zanata init to select a project without project type
- [ ] select a version with different project type compared to its project
- [ ] select a project and version both don't have project type

definite:
:+1:

PR #48, opened 2015-02-24
rhbz1156236 - deprecate locales in config file and fetch from server

huangp:
This depends on https://github.com/zanata/zanata-server/pull/651

### DIFF
--- a/zanata-cli/src/test/java/org/zanata/client/ZanataClientTest.java
+++ b/zanata-cli/src/test/java/org/zanata/client/ZanataClientTest.java
@@ -3,6 +3,7 @@ package org.zanata.client;
 import java.io.StringWriter;
 import java.net.URL;
 
+import org.hamcrest.CoreMatchers;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
@@ -183,6 +184,7 @@ public class ZanataClientTest {
         client.processArgs(command, "--url", url, "--project", project,
                 "--project-version", version);
 
+        assertThat(err.toString(), CoreMatchers.is(""));
         verify(mockOptions).setUrl(new URL(url));
         verify(mockOptions).setProj(project);
         verify(mockOptions).setProjectVersion(version);
--- a/zanata-client-commands/src/main/resources/prompts.properties
+++ b/zanata-client-commands/src/main/resources/prompts.properties
@@ -90,4 +90,4 @@ invalid.rule=Rule defined is not valid (must contain locale): %s
 unrecognized.variables=Valid variables are %s. Anything else in '%s' will be treated as literal value.
 confirm.rule=Continue with current mapping rule(s) (y/n)?
 
-locales.in.config.deprecated=Locale mappings in project config file (zanata.xml) is deprecated! Please remove it from the file so that it can use latest language settins on the server.
+locales.in.config.deprecated=Locale mappings are now handled using locale aliases on the server, so locale mappings in the project config file (zanata.xml) are now deprecated.\nPlease add a locale alias in the project language settings to replace each locale mapping in zanata.xml, then remove the <locales> section from zanata.xml.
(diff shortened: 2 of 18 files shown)


huangp:
@zanata-jenkins retest this please

huangp:
@davidmason can you review this please?

zanata-cli/src/test/java/org/zanata/client/ZanataClientTest.java
```diff
@@ -183,6 +184,7 @@ public void testStatsCommand() throws Exception {
         client.processArgs(command, "--url", url, "--project", project,
                 "--project-version", version);
 
+        assertThat(err.toString(), CoreMatchers.is(""));
```
davidmason: :thought_balloon: (minor) could `import static org.hamcrest.CoreMatchers.is`.

zanata-client-commands/src/main/resources/prompts.properties
```diff
@@ -89,3 +89,5 @@ no.default.mapping=Can not find default translation file mapping rule for projec
 invalid.rule=Rule defined is not valid (must contain locale): %s
 unrecognized.variables=Valid variables are %s. Anything else in '%s' will be treated as literal value.
 confirm.rule=Continue with current mapping rule(s) (y/n)?
+
+locales.in.config.deprecated=Locale mappings in project config file (zanata.xml) is deprecated! Please remove it from the file so that it can use latest language settins on the server.
```
davidmason: :speech_balloon: "Locale mappings in project config file (zanata.xml) are deprecated! Please set locale mappings in the server, then remove the <locales> section from the config file."

We could provide a link to the project/version language settings page where they would enter the settings.

davidmason: I will update the message. The key information it should include is:

 - locale mappings are now handled on the server
 - you should add a locale alias on the server for each locale mapping in zanata.xml
 - after adding aliases, you should remove <locales> from zanata.xml so that server locales will be used.

davidmason:
I am setting this to QA since nothing significant is present from code review.

djansen-redhat:
:small_red_triangle: If client attempts to push a language that is not in the local zanata.xml locales list, but *is* on the server, it will fail with client.exceptions.ConfigException

davidmason:
@djansen-redhat 

> :small_red_triangle: I can't tell if it's this PR or not, but nothing I've tried to push for a File type project has behaved correctly.
> Push type src does nothing

My first thought is that the command must have file types specified (txt, etc.), and with no file types it will not proceed. Could that be it?

carlosmunoz:
@djansen-redhat do you mean `push type source`?

huangp:
for point 1: That's a decision we need to make. Locales in zanata.xml will take precedence over locales on the server. So there is no need to fetch from the server if user has defined locales in file. Unless we want to keep two copies of the locale list (one for server and one for file), then when user provide locales as command line option, we would know whether it is actually valid. Personally I think this is a rare use case not worth implementing. Hopefully people will switch to server side locale config soon and this won't be a problem anymore.

djansen-redhat:
:small_red_triangle: Cannot push translations for a language "not in zanata.xml", despite not having a locales section in the xml

djansen-redhat:
@carlosmunoz yeah... that's what I *meant to* do...

djansen-redhat:
:white_check_mark: Tested.
We should improve the responses for the actions, as they are either vague or simply *not correct*.
e.g.
> Skipping file trans/en-AU.po; no locale entry found in zanata.xml
*a.* I don't have locales in my zanata.xml and *b.* I specified a locale other than the one above - I don't care! or *c.* if I did say en-AU but the server has an alias then the client uses that and still reports the warning.
The client throws noise at me that makes me uncomfortable.

djansen-redhat:
Did this pass review?

PR #49, opened 2015-03-02
Include licence (LGPL 2.1) in dist archives

seanf:


carlosmunoz:
:+1: Looks good.
A couple of failing builds though

davidmason:
:+1: All looks good to me too.

seanf:
The failing build is caused by checkstyle checks, which are currently checking against `master`'s newer checkstyle rules, but using `release`'s older version for the checkstyle plugin.  I think we need a way of keeping `release`'s checkstyle rules in sync with the version, either at the branch level (`release` vs `master`) or better still at the tag level (in this case, `zanata-parent-20`).  

The correct solution would probably involve `zanata-parent` exporting some files (including `checkstyle.xml` and `checkstyle-suppressions.xml`) to be used in the child modules.  Perhaps something like this: http://javasplitter.blogspot.com.au/2011/08/sharing-configuration-files-from-maven.html

In the meantime, I'm tempted to turn off checkstyle for the release branch.

definite:
:+1:

seanf:
Merged into `integration/master`: https://github.com/zanata/zanata-client/commit/8422146f9184b43bc94ef4fe6378d3e072c1e41b

PR #53, opened 2015-03-19
add src-dir trans-dir section

huangp:
same as https://github.com/zanata/zanata.github.io/pull/20

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -38,8 +38,18 @@ These steps should be repeated for each project-version before using any zanata-
 
 You can customize `zanata.xml` with command hooks so that other tools will automatically run before or after Zanata commands. Read about command hooks at the [command hook page on the wiki](https://github.com/zanata/zanata-server/wiki/Client-Command-Hooks).
 
+## Source directory and translation directory
+
+To prevent the need to specify the source and translation directories on the command-line, they can be specified in zanata.xml. The source and translation directories are specified in the `<src-dir>` and `<trans-dir>` elements respectively, as shown below. Both paths must be relative to the directory that contains zanata.xml.
+
+```xml
+<src-dir>po</src-dir>
+<trans-dir>.</trans-dir>
+```
 
 ## Locale Configuration
+
+***Note:*** Locale configuration in zanata.xml is being phased out, and this configuration will be specified on the server instead. When your project is on a server version that allows locale configuration and you have a client version that supports server locales, you should remove this configuration. Check the details in [feature request 1156236](https://bugzilla.redhat.com/show_bug.cgi?id=1156236).
 
 The `zanata.xml` will contain a list of locales so that the client knows which locales to push and pull to/from the Zanata server. When downloaded from the Zanata server, the list will have the locales as specified by the server itself. It will look something like this:
 


docs/configuration.md
```diff
@@ -38,9 +38,21 @@ These steps should be repeated for each project-version before using any zanata-
 
 You can customize `zanata.xml` with command hooks so that other tools will automatically run before or after Zanata commands. Read about command hooks at the [command hook page on the wiki](https://github.com/zanata/zanata-server/wiki/Client-Command-Hooks).
 
+## Source directory and translation directory
+
+`<src-dir>` and `<trans-dir>` can be defined in zanata.xml. You should use a relative path relative to the directory where zanata.xml lives. This will save you from giving the option on command line each time.
```
davidmason: I think we should flip this so the reason to do it is first:

> To prevent the need to specify the source and translation directories on the command-line, they can be specified in zanata.xml. The source and translation directories are specified in the `<src-dir>` and `<trans-dir>` elements respectively, as shown below. Both paths must be relative to the directory that contains zanata.xml.

docs/configuration.md
```diff
@@ -38,9 +38,21 @@ These steps should be repeated for each project-version before using any zanata-
...
+Example:
+
+```xml
+<src-dir>po</src-dir>
+<trans-dir>.</trans-dir>
+```
 
 ## Locale Configuration
 
+***NOTE:*** Once feature [1156236](https://bugzilla.redhat.com/show_bug.cgi?id=1156236) is released, you don't need to define any locales in your zanata.xml anymore.
```
davidmason: Can we just say which server version it will be supported in? We can mention to look at the bottom of any page on the server for the version?

> **Note:** Locale configuration in zanata.xml is being phased out. When your project is hosted on a server of version X.X or above, you should remove this configuration. The server version is shown at the bottom of any page on the server.

huangp: It's a little bit complicated in this case. We need to mention both server and client version which means users need to check both as well. This is a temporary notice anyway. Once we release both the whole section will need to change again so I'd like to keep it simple.

davidmason: Ok, how about:

> **Note:** Locale configuration in zanata.xml is being phased out, and this configuration will be specified on the server instead. When your project is on a server version that allows locale configuration and you have a client version that supports server locales, you should remove this configuration. Check the details in [feature request 1156236](https://bugzilla.redhat.com/show_bug.cgi?id=1156236).

docs/configuration.md
```diff
@@ -38,9 +38,19 @@ These steps should be repeated for each project-version before using any zanata-
...
+To prevent the need to specify the source and translation directories on the command-line, they can be specified in zanata.xml. The source and translation directories are specified in the `<src-dir>` and `<trans-dir>` elements respectively, as shown below. Both paths must be relative to the directory that contains zanata.xml.
+
+```xml
+<src-dir>po</src-dir>
+<trans-dir>.</trans-dir>
+```
 
 ## Locale Configuration
 
+***Note:*** Locale configuration in zanata.xml is being phased out, and this configuration will be specified on the server instead. When your project is on a server version that allows locale configuration and you have a client version that supports server locales, you should remove this configuration. Check the details in [this feature request](https://bugzilla.redhat.com/show_bug.cgi?id=1156236).
```
davidmason: > [this feature request]

I have read that it is better to make link text have meaning on its own, which is better for accessibility. I hear that visually impaired users have a terrible time with those blog posts that have links like "see [here](),  [here]() and [here]()".

So in this case I prefer to use something like "feature request 1156236" or "feature request to remove locale configuration in zanata.xml" to help users have a bit of an idea where they might be navigating.

davidmason:
@huangp I am passing this review, nothing worth holding it up on. Please take a look at my extra comment though since I think it is something good to consider in general.

PR #57, opened 2015-04-09
Implement doc type support for file project type

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1186972

1. common: https://github.com/zanata/zanata-common/pull/9
2. api: https://github.com/zanata/zanata-api/pull/26
3. **client**: https://github.com/zanata/zanata-client/pull/57
4. server: https://github.com/zanata/zanata-server/pull/772


stub-server/src/main/java/org/zanata/rest/service/MockFileResource.java
```diff
@@ -51,14 +62,27 @@
...
+        }
+        for (DocumentType docType: ProjectType.getSupportedSourceFileTypes(
+            ProjectType.Gettext)) {
+            extensions.addAll(docType.getSourceExtensions());
+        }
         StringSet entity = new StringSet("");
-        entity.addAll(ProjectType.getSupportedSourceFileTypes(ProjectType.File));
-        entity.addAll(ProjectType
-                .getSupportedSourceFileTypes(ProjectType.Gettext));
+        entity.addAll(extensions);
```
davidmason: What is the reason for the intermediate `Set<String> extensions`, instead of just adding everything directly to `StringSet entity`?

zanata-client-commands/src/main/java/org/zanata/client/commands/FileMappingRuleHandler.java
```diff
@@ -108,9 +116,9 @@ private boolean matchFileExtensionWithProjectType(
      */
     public String getRelativeTransFilePathForSourceDoc(
             QualifiedSrcDocName qualifiedSrcDocName,
-            @Nonnull LocaleMapping localeMapping) {
+            @Nonnull LocaleMapping localeMapping, Optional<String> transExt) {
```
davidmason: Please give `transExt` a more meaningful name. Is it the translation file extension?

zanata-client-commands/src/main/java/org/zanata/client/commands/FileMappingRuleHandler.java
```diff
@@ -125,11 +133,15 @@ public String getRelativeTransFilePathForSourceDoc(
 
     @VisibleForTesting
     protected static EnumMap<Placeholders, String> parseToMap(
-            @Nonnull String sourceFile, @Nonnull LocaleMapping localeMapping) {
+            @Nonnull String sourceFile, @Nonnull LocaleMapping localeMapping,
+            Optional<String> transExt) {
```
davidmason: - [ ] Please give `transExt` a more meaningful name here.

zanata-client-commands/src/main/java/org/zanata/client/commands/TransFileResolver.java
```diff
@@ -80,16 +80,16 @@ public TransFileResolver(ConfigurableProjectOptions opts) {
      * @return translation destination
      */
     public File resolveTransFile(QualifiedSrcDocName qualifiedSrcDocName,
-            LocaleMapping localeMapping) {
+            LocaleMapping localeMapping, Optional<String> transExt) {
```
davidmason: - [ ] Please give `transExt` a more meaningful name here.

zanata-client-commands/src/main/java/org/zanata/client/commands/TransFileResolver.java
```diff
@@ -121,17 +122,18 @@ private ProjectType getProjectType() {
 
     private File getDefaultTransFileFromProjectType(
             QualifiedSrcDocName qualifiedSrcDocName, LocaleMapping localeMapping,
-            ProjectType projectType) {
+            ProjectType projectType, Optional<String> transExt) {
```
davidmason: - [ ] Please give `transExt` a more meaningful name here.

zanata-client-commands/src/main/java/org/zanata/client/commands/TransFileResolver.java
```diff
@@ -121,17 +122,18 @@ private ProjectType getProjectType() {
...
                 .getRelativeTransFilePathForSourceDoc(qualifiedSrcDocName,
-                        localeMapping);
+                        localeMapping, transExt);
         return new File(opts.getTransDir(), relativePath);
     }
 
     private Optional<File> tryGetTransFileFromProjectMappingRules(
-            QualifiedSrcDocName qualifiedSrcDocName, LocaleMapping localeMapping) {
+            QualifiedSrcDocName qualifiedSrcDocName, LocaleMapping localeMapping,
+            Optional<String> transExt) {
```
davidmason: - [ ] Please give `transExt` a more meaningful name here.

zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PropertiesStrategy.java
```diff
@@ -63,11 +63,13 @@ public FileDetails writeTransFile(Resource doc, String docName,
...
-                    localeMapping.getJavaLocale(), true);
+            PropWriter.writeTranslations(doc, targetDoc,
+                getOpts().getTransDir(), docName,
+                localeMapping.getJavaLocale(), PropWriter.CHARSET.Latin1, true);
         } else {
-            PropWriter.write(null, targetDoc, getOpts().getTransDir(), docName,
-                    localeMapping.getJavaLocale(), false);
+            PropWriter.writeTranslations(null, targetDoc,
+                getOpts().getTransDir(), docName,
+                localeMapping.getJavaLocale(), PropWriter.CHARSET.Latin1, false);
```
davidmason: Could probably simplify this to:

```
PropWriter.writeTranslations(
    createSkeletons ? doc : null,
    targetDoc,
    getOpts().getTransDir(),
    docName,
    localeMapping.getJavaLocale(),
    PropWriter.CHARSET.Latin1, 
    createSkeletons);
```

zanata-client-commands/src/main/java/org/zanata/client/commands/pull/RawPullCommand.java
```diff
@@ -201,8 +206,17 @@ public void run() throws IOException {
...
                                 try {
-                                    strat.writeTransFile(localDocName, locMapping,
-                                            transDoc);
+                                    String fileName =
+                                            ClientUtil.getFileNameFromHeader(
+                                                    response.getHeaders());
+                                    String targetFileExt = FilenameUtils
+                                                    .getExtension(fileName);
+
+                                    Optional transExt =
```
davidmason: - [ ] please give `transExt` a more meaningful name.

zanata-client-commands/src/main/java/org/zanata/client/commands/pull/RawPullStrategy.java
```diff
@@ -62,15 +64,15 @@ public void writeSrcFile(String localDocName, InputStream srcFile)
     }
 
     public void writeTransFile(String localDocName,
-            LocaleMapping localeMapping, InputStream transFile)
+            LocaleMapping localeMapping, InputStream transFile, Optional<String> transExt)
```
davidmason: - [ ] Please give `transExt` a more meaningful name.

zanata-client-commands/src/main/java/org/zanata/client/commands/pull/UTF8PropertiesStrategy.java
```diff
@@ -51,11 +51,16 @@ public FileDetails writeTransFile(Resource doc, String docName,
...
+                docName, localeMapping.getJavaLocale(),
+                PropWriter.CHARSET.UTF8, true);
         } else {
-            PropWriter.writeUTF8(null, targetDoc, getOpts().getTransDir(),
-                    docName, localeMapping.getJavaLocale(), false);
+            PropWriter.writeTranslations(null, targetDoc,
+                getOpts().getTransDir(),
+                    docName, localeMapping.getJavaLocale(),
+                    PropWriter.CHARSET.UTF8,
+                false);
```
davidmason: - [ ] Can simplify:

```
PropWriter.writeTranslations(
    createSkeletons ? doc : null,
    targetDoc,
    getOpts().getTransDir(),
    docName,
    localeMapping.getJavaLocale(),
    PropWriter.CHARSET.UTF8,
    createSkeletons);
```

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+            return typeWithExtension.toUpperCase();
+        }
+    }
+
+    /**
+     * Extract extensions in square bracket
+     * input pattern: fileType[extension;extension], [extension;extension]
+     */
+    private List<String> extractExtensions(String typeWithExtension) {
+        Pattern p = Pattern.compile("\\[(.*?)\\]");
```
davidmason: We could just use a single regular expression for file type and extensions. The advantage would be that the regex describes the whole pattern we are expecting so it is easier to understand what is expected from the code:

Something like this (first group is the type, second group is the extensions):

```
private static final Pattern fileNameExtensionsPattern = Pattern.complie("([^[*])(?:[(.*?)])");
```

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+     */
+    private void validateInputFileType(String inputFileType,
+            List<String> inputExtensions, List<DocumentType> acceptedTypes) {
+        if (StringUtils.isNotBlank(inputFileType)) {
+            return;
+        }
+        if (inputExtensions.isEmpty()) {
+            //throw error if inputFileType and inputExtensions is empty
+            throw new RuntimeException(
+                    "Invalid expression for '--file-types' option");
```
davidmason: Would be nice to have a more specific exception here, like `InvalidUserInputException` or something.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+            for (DocumentType docType: acceptedTypes) {
+                for (String extension: docType.getSourceExtensions()) {
+                    if (inputExtensions.contains(extension)) {
+                        String msg = "Please use file type '"
+                            + docType
+                            + "' for extension '"
+                            + extension
+                            + "', e.g. --file-types "
+                            + docType;
+                        throw new RuntimeException(msg);
```
davidmason: This could use a more specific exception as well.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+            //suggest --file-types options for this extension
+            for (DocumentType docType: acceptedTypes) {
+                for (String extension: docType.getSourceExtensions()) {
+                    if (inputExtensions.contains(extension)) {
+                        String msg = "Please use file type '"
+                            + docType
+                            + "' for extension '"
+                            + extension
+                            + "', e.g. --file-types "
+                            + docType;
```
davidmason: - [ ] This string cannot be localised. It is intended to be shown to users, so it may be worth sorting out.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+                        throw new RuntimeException(msg);
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * Return list of validated type with list of extensions
```
davidmason: Return **map** of validated type with **set** of extensions

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+     *
+     * Validate user input file types against server accepted file types
+     *
+     * - Each file type is only input once
+     * - Same file extension cannot be in multiple file type
+     * - Source file extension that is not supported from file type will be ignored.
```
davidmason: - [ ] Just ignored, or generate a warning? We should at least warn the user that we are ignoring something they input (and if we are, this comment should say so).

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+     *
+     * eg. input pattern --file-types properties_utf8,properties[txt],plain_text[md;markdown],html[txt],idml
```
davidmason: - [ ] It looks like this is an example of a invalid input (due to use of `txt` twice). Better make it clear what the example is supposed to show.
- [ ] Even better, show an example of valid input, and an example of invalid input for each of the 3 rules above. The same examples can be copied to the help output and documentation for this option.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+            validateInputFileType(fileType, extensions, acceptedTypes);
+
+            DocumentType inputFileType = DocumentType.getByName(fileType);
+
+            //skip file type if its not supported
+            if (inputFileType == null) {
+                String msg =
+                        "Requested file type '"
+                                + fileType
+                                + "' is not supported by the target server and will be ignored.";
```
davidmason: - [ ] String cannot be localised.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+                consoleInteractor.printfln(DisplayMode.Warning, msg);
+                continue;
+            }
+
+            //throw error if file type is input more than once
+            if (filteredFileTypes.containsKey(inputFileType)) {
+                String msg =
+                    "File type '" + fileType + "' cannot be used more than once in option --file-types.";
```
davidmason: - [ ] String cannot be localised.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+                //throw exception if user input extension if not part of DocumentType supported source extension
+                for (String extension: extensions) {
+                    if (!inputFileType.getSourceExtensions()
+                            .contains(extension)) {
+                        String msg =
+                                "Extension '"
+                                        + extension
+                                        + "' for file type '"
+                                        + fileType
+                                        + "' is not supported by the target server and will be ignored.";
```
davidmason: - [ ] String cannot be localised.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+            if (filteredFileTypes.containsKey(inputFileType)) {
+                String msg =
+                    "File type '" + fileType + "' cannot be used more than once in option --file-types.";
+                log.error(msg);
+                throw new RuntimeException(msg);
+            }
+
+            Set<String> filteredExtensions = new HashSet<>(extensions);
+
+            if (!extensions.isEmpty()) {
```
davidmason: - [ ] Extract this whole block to a method, like `validateExtensionsMatchFiletype`

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+                //throw exception if user input extension if not part of DocumentType supported source extension
+                for (String extension: extensions) {
+                    if (!inputFileType.getSourceExtensions()
+                            .contains(extension)) {
```
davidmason: This does not match what we discussed. This stops someone from uploading markdown as plain text. The user should be able to use any extension they want to match a file type.

e.g. user inputs `... --file-types plain_text[md;markdown]`

It looks like this code would make sure it will never upload any markdown document because "md" and "markdown" are not part of the plain_text source extensions.

- [ ] We should remove this check, and just make sure the client sends the `DocumentType` when it sends the file. For old clients that do not send DocumentType, the server can still just use the extension of the uploaded file (since old clients will only send extensions that it recognizes).

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+                }
+            }
+
+            /**
+             * Use the extensions from typeWithExtension input if exists,
+             * otherwise, use the extensions from server.
+             */
+            filteredExtensions =
+                filteredExtensions.isEmpty() ? inputFileType.getSourceExtensions()
+                            : filteredExtensions;
```
davidmason: I don't think we should be running the command using default extensions unless they actually asked for them (by not including any extensions). If something they put in is not valid, I think it is better to just fail the command and tell them about it so they can fix it, so they have a chance to make it work the way they intended.

davidmason: To clarify, I think this line is fine, but it should not reach this line if they have entered something invalid.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+            //throw error if same file extension found in multiple file type
+            for (DocumentType filteredDocType: filteredFileTypes.keySet()) {
+                for (String filteredExtension: filteredDocType.getSourceExtensions()) {
+                    if (filteredExtensions.contains(filteredExtension)) {
+                        String msg =
+                                "Conflict file type between '" + fileType
+                                        + "' and '" + filteredDocType.name() + "' "
+                                        + "for extension '"
+                                        + filteredExtension
+                                        + "' in option --file-types.";
```
davidmason: - [ ] String cannot be localised.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +94,164 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+        }
+        return filteredFileTypes;
+    }
+
     @Override
     public void run() throws IOException {
         PushCommand.logOptions(log, getOpts());
-        log.warn("Using EXPERIMENTAL project type 'file'.");
+        consoleInteractor.printfln(DisplayMode.Warning,
+            "Using EXPERIMENTAL project type 'file'.");
```
davidmason: We should think about when to stop calling this experimental.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -98,37 +273,34 @@ public void run() throws IOException {
...
+            validateFileTypes(serverAcceptedTypes, getOpts().getFileTypes());
 
-        if (types.isEmpty()) {
+        if (filteredDocTypes.isEmpty()) {
             log.info("no valid types specified; nothing to do");
             return;
         }
 
+        ImmutableList.Builder<String> sourceFileExtensionsBuilder = ImmutableList.builder();
+        for (Set<String> filteredSourceExtensions: filteredDocTypes.values()) {
```
davidmason: - [ ] (style) add space before colon: `...urceExtensions : filteredDo...`

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -151,12 +323,17 @@ public void run() throws IOException {
...
-            log.info("Found source documents:");
+            consoleInteractor.printfln(DisplayMode.Information,
+                "Found source documents:");
             for (String docName : localDocNames) {
                 if (docsToPush.contains(docName)) {
-                    log.info("           {}", docName);
+                    DocumentType fileType = getFileType(filteredDocTypes,
+                        getExtensionFor(docName));
+                    consoleInteractor.printfln(DisplayMode.Information,
+                        "           " + docName + " [" + fileType.name() + "]");
```
davidmason: - [ ] String cannot be localised.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -151,12 +323,17 @@ public void run() throws IOException {
...
                 } else {
-                    log.info("(to skip)  {}", docName);
+                    consoleInteractor.printfln(DisplayMode.Information,
+                        "(to skip)  " + docName);
```
davidmason: - [ ] String cannot be localised.

From the look of the method name, `printfln(...)` probably allows placeholders, please use them.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -167,9 +344,10 @@ public void run() throws IOException {
                 throw new ConfigException("pushType set to '"
                         + getOpts().getPushType()
                         + "', but project has no locales configured");
-            log.warn("pushType set to '"
-                    + getOpts().getPushType()
-                    + "': existing translations on server may be overwritten/deleted");
+            consoleInteractor.printfln(DisplayMode.Warning, "pushType set to '"
+                + getOpts().getPushType()
+                +
+                "': existing translations on server may be overwritten/deleted");
```
davidmason: - [ ] String cannot be localised.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -186,8 +364,12 @@ public void run() throws IOException {
 
         for (final String localDocName : docsToPush) {
             try {
-                final String fileType = getExtensionFor(localDocName);
+                final String srcExtension = getExtensionFor(localDocName);
+                final DocumentType fileType = getFileType(filteredDocTypes,
+                    srcExtension);
                 final String qualifiedDocName = qualifiedDocName(localDocName);
+                Optional transExt = Optional.fromNullable(
```
davidmason: - [ ] Please give `transExt` a more meaningful name.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -241,6 +423,25 @@ public void visit(LocaleMapping locale,
     }
 
     /**
+     * Search and return file type with given source file extension
+     *
+     * @param fileTypes
+     * @param fileExtension
+     * @return
```
davidmason: - [ ] Either say what it returns, or delete the `@return` line.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -241,6 +423,25 @@ public void visit(LocaleMapping locale,
...
+     */
+    private @Nullable DocumentType getFileType(Map<DocumentType, Set<String>> fileTypes,
```
davidmason: - [ ] Is there a method in `DocumentType` that can do this? Or is `fileTypes` a narrower set of types? Maybe this should go into DocumentType anyway (it would just need a `Set<DocumentType>` and `String` since it already knows which source extensions are valid for each).

zanata-client-commands/src/test/java/org/zanata/client/MockServerRule.java
```diff
@@ -288,10 +287,13 @@ public PullCommand createPullCommand(List<ResourceMeta> remoteDocList,
     public RawPushCommand createRawPushCommand() {
         when(clientFactory.getFileResourceClient()).thenReturn(
                 fileResourceClient);
-        List<String> fileTypes =
-                ProjectType.getSupportedSourceFileTypes(ProjectType.File);
-        when(fileResourceClient.acceptedFileTypes())
-                .thenReturn(new StringSet(Joiner.on(";").join(fileTypes)));
+
+        List<DocumentType> fileProjectDocTypes =
```
davidmason: - [ ] The name is a little confusing, how about `documentTypes`?

zanata-rest-client/src/main/java/org/zanata/rest/client/ClientUtil.java
```diff
@@ -107,4 +110,25 @@ public static String getBaseURL(String movedTo) {
...
+     * Extract filename from response header.
+     *
+     * e.g. Content-Disposition=[attachment; filename="RPM.po"]
+     *
+     * @param headers
+     * @return
+     */
+    public static String getFileNameFromHeader(MultivaluedMap<String, String> headers) {
+        final String CONTENT_DISPOSITION_HEADER = "Content-Disposition";
+        final String FILENAME_PATTERN = "filename=\"(.*?)\"";
```
davidmason: - [ ] Can the filename ever contain a double-quote character? If so this will fail for an escaped quote character because it uses lazy repeat (`*?`). Please figure out whether there can be a quote, and if so how it can be escaped. I can help make a regex that will match it accurately.

zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java
```diff
@@ -50,14 +55,14 @@
...
                 .resource(baseUri)
-                .path(FileResource.SERVICE_PATH)
-                .path("accepted_types")
-                .accept(MediaType.TEXT_PLAIN_TYPE)
-                .get(String.class);
-        return new StringSet(types);
+                        .path(FileResource.SERVICE_PATH
+                                + FileResource.ACCEPTED_TYPE_LIST_RESOURCE)
+            .get(new GenericType<List<DocumentType>>() {
+            });
```
davidmason: - [ ] Something strange happened to indentation here. Line 60, 61 and 63 should all start at the same column.

zanata-rest-client/src/test/java/org/zanata/rest/client/FileResourceClientTest.java
```diff
@@ -67,12 +74,16 @@ public void setUp() throws URISyntaxException {
...
 
-        assertThat(serverAcceptedTypes, Matchers.containsInAnyOrder("dtd",
+        Set<String> allExtension = new HashSet<String>();
+        for (DocumentType docType : serverAcceptedTypes) {
+            allExtension.addAll(docType.getSourceExtensions());
+        }
+        assertThat(allExtension, Matchers.containsInAnyOrder("dtd",
                 "txt", "idml", "html", "htm", "odt", "odp", "odg", "ods",
-                "srt", "sbt", "sub", "vtt", "pot"));
+                "srt", "sbt", "sub", "vtt", "properties", "xml"));
```
davidmason: - [ ] What happened to `"pot"`?

stub-server/src/main/java/org/zanata/rest/service/MockFileResource.java
aeng: To prevent human error in duplication.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
aeng: Not sure what you meant by 'sorting out'.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
aeng: Yup.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
aeng: The problem is that we have mapping and source and translation file extension from server side. If we allow customised extension file format, how are we gonna match up with the translation file? e.g PO and POT.

zanata-rest-client/src/main/java/org/zanata/rest/client/ClientUtil.java
aeng: Yes. File name can contain double-quote.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
aeng: If user customised the extensions, then it will need the `Map<DocumentType, Set<String>>`  and makes more sense to be in this class.

stub-server/src/main/java/org/zanata/rest/service/MockFileResource.java
davidmason: I don't understand what you mean here. Wouldn't less lines of code mean there is less chance of human error?

I can't figure out what the problem would be with using the `StringSet` directly, like this:

```
        StringSet extensions = new StringSet("");
        for (DocumentType docType: ProjectType.getSupportedSourceFileTypes(ProjectType.File)) {
            extensions.addAll(docType.getSourceExtensions());
        }
        for (DocumentType docType: ProjectType.getSupportedSourceFileTypes(
            ProjectType.Gettext)) {
            extensions.addAll(docType.getSourceExtensions());
        }
        return Response.ok(extensions.toString()).build();
```

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +96,174 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+    }
+
+    /**
+     * Extract input string to type and extensions
+     * input pattern: fileType[extension;extension], [extension;extension]
+     *
+     * @param typeWithExtension
+     * @return result[0] - type, result[1] - extensions. e.g [txt,html]
+     */
+    private String[] extractTypeAndExtension(String typeWithExtension) {
```
davidmason: - [ ] I was suggesting to use a single regex, but I think it is better to keep the two methods to get each of the parts (just use the same regex in them).

Java is too clunky to return multiple results like this without a class for the results. I was going to say it is not worth making a class just for this, but then it might make for cleaner code to have a class that represents 1 of the file type entries (type + extensions) - there are several other places such a class could be used.

stub-server/src/main/java/org/zanata/rest/service/MockFileResource.java
aeng: @davidmason my mistake. StringSet it is.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
aeng: Yeah. I was doubting myself if I should have a wrapper for type+extension. Maybe I should to make it cleaner.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
davidmason: > Not sure what you meant by 'sorting out'.

Sorry, I mean it may be worth making it so it can be localised. Whatever the standard way is that we make format strings in the client at the moment.

davidmason: Oh, now I see that you already did that. :+1: :confetti_ball:

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +96,174 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+    /**
+     * Return map of validated type with set of extensions
+     *
+     * Validate user input file types against server accepted file types
+     *
+     * Valid input - properties_utf8,properties[txt],plain_text[md;markdown]
+     *
+     * - Each file type is only input once - e.g. html,properties,html
+     * - Same file extension cannot be in multiple file type - e.g. plain_text[txt],properties[txt]
+     * - Generate warning and ignore for source file extension that is not supported from file type. - e.g idml[txt]
```
davidmason: - [ ] we can make it more clear whether the examples are showing valid or invalid inputs, like this:

```
   * - Each file type is only input once. e.g.
   *    - valid: "html,properties,txt"
   *    - invalid: "html,properties,html"
```

- [ ] It looks like this is an example of a invalid input (due to use of `txt` twice). Better make it clear what the example is supposed to show.
- [ ] Even better, show an example of valid input, and an example of invalid input for each of the 3 rules above. The same examples can be copied to the help output and documentation for this option.

davidmason: > Generate warning and ignore for source file extension that is not supported from file type.

- [ ] This is not good behaviour for users. They should be able to use whatever extension they want for their files. If they have text files that they have put their own custom extension on, this would force them to rename them all before they can upload. We should just use the extension they give us and not complain (as long as it is not a duplicate).

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
davidmason: If we don't know the extension, then we can just use the same extension for source and translation. That is all users will want most of the time.

For the rare case that someone does need a custom extension that is different between source and translation, it is ok for that case not to work. I think it will be possible later when we update the File project type to handle different directory structures.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
```diff
@@ -75,10 +96,174 @@ public RawPushCommand(PushOptions opts, RestClientFactory clientFactory) {
...
+            Set<String> filteredExtensions, DocumentType inputFileType) {
+        if (!extensions.isEmpty()) {
+            for (String extension: extensions) {
+                if (!inputFileType.getSourceExtensions()
+                    .contains(extension)) {
+                    String msg = Messages.format(
+                        "file.type.notSupportedExtension", extension,
+                        inputFileType.name());
+                    consoleInteractor.printfln(DisplayMode.Warning, msg);
+                    filteredExtensions.remove(extension);
```
davidmason: - [ ] We should accept custom extensions, as long as the user specified a file type so we know what type to send the file as.

zanata-client-commands/src/main/resources/prompts.properties
```diff
@@ -91,3 +91,21 @@ unrecognized.variables=Valid variables are %s. Anything else in '%s' will be tre
...
+# {0} - fileType, {1} - file extension, {2} - fileType
+file.type.suggestFromExtension=Please use file type '{0}' for extension '{1}', e.g --file-types {2}
+# {0} - fileType
+file.type.duplicateFileType=File type '{0}' cannot be used more than once in option --file-types.
+# {0} - file extension, {1} - file type
+file.type.notSupportedExtension=Extension '{0}' for file type '{1}' is not supported by the target server and will be ignored.
+# {0} - fileType
+file.type.typeNotSupported=Requested file type '{0}' is not supported by the target server and will be ignored.
+# {0} - fileType1, {1} - fileType2, {2} - extension
+file.type.conflictExtension=Conflict file type between '{0}' and '{1}' for extension '{2}' in option --file-types.
```
davidmason: - [ ] I think we can state this more plainly, like "You used extension '{2}' in both '{0}' and '{1}' in option --file-types. Extensions can only be used for one type."

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
davidmason: - [ ] @aeng it looks like you stopped calling it experimental. Do our QA/UX guys think it is stable and feature-complete enough for that?

aeng: Oh.. must have removed it by mistake. 
It should be still experimental.

Regards, 

Alex Eng 
Senior Software Engineer 
DID: +61 3514 8262 
Mobile: +614 2335 3457 
Email: aeng@redhat.com 

Red Hat, Asia-Pacific Pty Ltd 
Level 1, 193 North Quay 
Brisbane 4000 
Office: +61 7 3514 8100 
Fax: +61 7 3514 8199 
Website: www.redhat.com 

----- Original Message -----

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
davidmason: Oh yeah, I didn't think of that :confused::thought_balloon: . Makes sense :+1:.

zanata-client-commands/src/main/java/org/zanata/client/commands/push/RawPushCommand.java
davidmason: Yeah, I think just having something like the following. It is not the most efficient, but it won't make a measurable difference to performance:

```
private static final Pattern fileNameExtensionsPattern = Pattern.compile("(.*)?(\\[(.*?)\\])?");

private String extractType(String typeWithExtension) {
    // run regex and extract group 1
}

private String extractExtensions(String typeWithExtension) {
    // run regex and extract group 2
}
```

davidmason:
The only remaining item is whether to allow users to use non-standard extensions, such as using `.markdown` for plain-text. If we allow it users will have a better experience.

It is a detail we missed in the initial requirements of the feature, but it is so great for usability that we should take the opportunity to allow it.

davidmason:
:+1:

definite:
The project Spacewalk failed on:
java/code/src/com/redhat/rhn/frontend/strings/jsp/StringResource.xml

Project URL:  https://github.com/spacewalkproject/spacewalk.git/

Server logs:
Caused by: com.mysql.jdbc.PacketTooBigException: Packet for query is too large (1068490 > 1048576). You can change this value on the server by setting the max_allowed_packet' variable.
	at com.mysql.jdbc.MysqlIO.send(MysqlIO.java:3279)
	at com.mysql.jdbc.MysqlIO.sendCommand(MysqlIO.java:1971)
	at com.mysql.jdbc.MysqlIO.sqlQueryDirect(MysqlIO.java:2151)
	at com.mysql.jdbc.ConnectionImpl.execSQL(ConnectionImpl.java:2625)
	at com.mysql.jdbc.PreparedStatement.executeInternal(PreparedStatement.java:2119)
	at com.mysql.jdbc.PreparedStatement.executeUpdate(PreparedStatement.java:2415)
	at com.mysql.jdbc.PreparedStatement.executeUpdate(PreparedStatement.java:2333)
	at com.mysql.jdbc.PreparedStatement.executeUpdate(PreparedStatement.java:2318)
	at sun.reflect.GeneratedMethodAccessor572.invoke(Unknown Source) [:1.7.0_79]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) [rt.jar:1.7.0_79]
	at java.lang.reflect.Method.invoke(Method.java:606) [rt.jar:1.7.0_79]
	at net.bull.javamelody.JdbcWrapper.doExecute(JdbcWrapper.java:403) [javamelody-core-1.52.0.jar:1.52.0]
	at net.bull.javamelody.JdbcWrapper$StatementInvocationHandler.invoke(JdbcWrapper.java:128) [javamelody-core-1.52.0.jar:1.52.0]
	at net.bull.javamelody.JdbcWrapper$DelegatingInvocationHandler.invoke(JdbcWrapper.java:285) [javamelody-core-1.52.0.jar:1.52.0]
	at com.sun.proxy.$Proxy190.executeUpdate(Unknown Source)
	at org.jboss.jca.adapters.jdbc.WrappedPreparedStatement.executeUpdate(WrappedPreparedStatement.java:493)
	at org.hibernate.engine.jdbc.internal.ResultSetReturnImpl.executeUpdate(ResultSetReturnImpl.java:186) [hibernate-core.jar:4.2.18.Final-redhat-2]
	... 171 more

definite:
zanata.xml


<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<config xmlns="http://zanata.org/namespace/config/">
  <url>$SERVER_URL</url>
  <project>spacewalk</project>
  <project-version>master</project-version>
  <project-type>file</project-type>
  <src-dir>.</src-dir>
  <trans-dir>.</trans-dir>
  <includes>**/StringResource*.xml,**/*.pot</includes>
  <excludes>**/StringResource_*.xml</excludes>
  <rules>
    <rule pattern="**/*.xml">{path}/{filename}_{locale_with_underscore}.xml</rule>
    <rule pattern="**/*.pot">{path}/{locale_with_underscore}.po</rule>
  </rules>  
</config>

aeng:
> It failed on big xliff document (1.1M):

@definite sounds like its a config issue with your database limit. See http://dev.mysql.com/doc/refman/5.1/en/server-system-variables.html#sysvar_max_allowed_packet. The default value should be 1GB.

definite:
@aeng  The empty translation should not be pulled unless --create-skeletons is specified.

aeng:
@definite The existing File-project-type implementation doesn't use that parameter from client. It will always set true to create-skeleton. If this require to change, it should be a separate request from the PR.

definite:
Chunk size issues should be addressed in https://bugzilla.redhat.com/show_bug.cgi?id=1220226

aeng:
Unhandle exception for multiple chunk upload - https://bugzilla.redhat.com/show_bug.cgi?id=1220230

definite:
"Empty translation should not be pulled" is addressed in https://bugzilla.redhat.com/show_bug.cgi?id=1220239

definite:
The bugs I found are not in the scope of this PR. 
So this PR is verified.

aeng:
@zanata-jenkins retest this please.

aeng:
Merged into integration/master. https://github.com/zanata/zanata-client/commit/57b61850cc35e0f01182531062fe74a55c698400

PR #58, opened 2015-04-23
rely on content not content state when determine what can be omitted in push

huangp:
When dealing with plural entries, the content state is not always reliable.

### DIFF
--- a/zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java
+++ b/zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java
@@ -39,9 +39,12 @@ import org.zanata.rest.dto.resource.TextFlowTarget;
 import org.zanata.rest.dto.resource.TranslationsResource;
 
 import com.google.common.base.Predicate;
-import com.google.common.collect.Collections2;
+import com.google.common.base.Strings;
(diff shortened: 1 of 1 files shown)


zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java
```diff
@@ -452,14 +455,25 @@ private static void stripUntranslatedEntriesIfMergeTypeIsNotImport(
...
                                 @Override
                                 public boolean apply(
                                         TextFlowTarget input) {
-                                    return input == null ||
-                                            input.getState().isUntranslated();
+                                    // it's unsafe to rely on content state (plural entries)
+                                    return input == null
+                                            || input.getContents().isEmpty()
+                                            || all(input.getContents(),
+                                                    blankStringPredicate);
```
aeng: what about the `input.getState().isUntranslated();` condition?

huangp: It's not reliable when dealing with plural entries. We treat it as untranslated if not all plurals have value. Therefore the original implementation will miss some partial translations

aeng:
:+1:

PR #60, opened 2015-04-28
rhbz903964 - try to display better error message if available

huangp:
see https://github.com/zanata/zanata-server/pull/787


definite:
:+1:

PR #62, opened 2015-05-15
update deprecated reference used

aeng:


zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
```diff
@@ -89,13 +92,8 @@ public PushPullCommand(O opts) {
...
     protected void confirmWithUser(String message) throws IOException {
         if (getOpts().isInteractiveMode()) {
-            Console console = System.console();
-            if (console == null) {
-                throw new RuntimeException(
-                        "console not available: please run Maven from a console, or use batch mode option (-B)");
-            }
-            console.printf(message + "\nAre you sure (y/n)? ");
-            expectYes(console);
+            console.printf(Question, message + "\nAre you sure (y/n)? ");
```
seanf: `ConsoleInteractorImpl` doesn't appear to handle the case where `System.console()` is null (when running with `nohup` or something?), so I think this introduces a danger of NPE.

aeng: The check should be in ConsoleInteractorImpl() constructor.

huangp:
:+1:

seanf:
:+1:

aeng:
Affected behaviour:

- [ ] Pull statistics and output as CSV
- [ ] Raw file project type push (make sure the console output text properly)
- [ ] Push pull command (other than file project type action, make sure the console output text properly)

aeng:
Merged into integration/master. https://github.com/zanata/zanata-client/commit/cde28583f5648ea65790ad1366c974d625371b24

stub-server/src/main/java/org/zanata/rest/service/MockFileResource.java
```diff
@@ -61,6 +61,7 @@
 @Path(FileResource.SERVICE_PATH)
 public class MockFileResource implements FileResource {
     @Override
+    @Deprecated
```
seanf: This should have an `@deprecated` tag in the javadoc pointing to an undeprecated alternative.

aeng: There's already JavaDoc in the parent class `FileResource`. I'm not sure if we should update all method that implements `FileResource`.

seanf: Well, if `FileResource` has a javadoc with a suitable `@deprecated` tag, that can be inherited (as long as the children don't replace the parent javadoc).

PR #63, opened 2015-05-19
rhbz1215274 - Min doc percent

huangp:
https://bugzilla.redhat.com/show_bug.cgi?id=1215274

### DIFF
--- a/zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
+++ b/zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
@@ -27,6 +27,7 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.StringWriter;
+import java.math.BigInteger;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
@@ -257,9 +258,10 @@ public abstract class PushPullCommand<O extends PushPullOptions> extends
         }
     }
 
-    protected Map<String, Map<LocaleId, TranslatedPercent>> getDocsTranslatedPercent() {
+    protected Map<String, Map<LocaleId, TranslatedPercent>> getDocsTranslatedPercent(
+            LocaleList locales) {
         ContainerTranslationStatistics statistics =
-                getDetailStatisticsForProjectVersion();
+                getDetailStatisticsForProjectVersion(locales);
         List<ContainerTranslationStatistics> statsPerDoc =
                 statistics.getDetailedStats();
         ImmutableMap.Builder<String, Map<LocaleId, TranslatedPercent>> docIdToStatsBuilder =
@@ -289,10 +291,15 @@ public abstract class PushPullCommand<O extends PushPullOptions> extends
     }
 
     @VisibleForTesting
-    protected ContainerTranslationStatistics getDetailStatisticsForProjectVersion() {
+    protected ContainerTranslationStatistics getDetailStatisticsForProjectVersion(
+            LocaleList locales) {
+        String[] localesOnServer = new String[locales.size()];
+        for (int i = 0; i < locales.size(); i++) {
+             localesOnServer[i] = locales.get(i).getLocale();
+        }
         return statsClient
                     .getStatistics(getOpts().getProj(),
-                            getOpts().getProjectVersion(), true, false, null);
+                            getOpts().getProjectVersion(), true, false, localesOnServer);
     }
 
     /**
@@ -304,17 +311,18 @@ public abstract class PushPullCommand<O extends PushPullOptions> extends
      * Optional.absence().
      *
      * @param pullTarget whether we need to pull translation target
+     * @param locales
      * @return either detailed document statistics or optional.absence()
      */
     protected Optional<Map<String, Map<LocaleId, TranslatedPercent>>> prepareStatsIfApplicable(
-            boolean pullTarget) {
+            boolean pullTarget, LocaleList locales) {
 
         Optional<Map<String, Map<LocaleId, TranslatedPercent>>> optionalStats =
                 Optional.absent();
         O opts = getOpts();
         if (pullTarget && opts instanceof PullOptions &&
                 ((PullOptions) opts).getMinDocPercent() > 0) {
-            optionalStats = Optional.of(getDocsTranslatedPercent());
+            optionalStats = Optional.of(getDocsTranslatedPercent(locales));
         }
         return optionalStats;
     }
@@ -335,7 +343,7 @@ public abstract class PushPullCommand<O extends PushPullOptions> extends
     }
 
     protected static class TranslatedPercent {
-        protected final int translatedPercent;
+        private final double translatedPercent;
         private final long total;
         private final long translated;
         private final long approved;
@@ -344,7 +352,14 @@ public abstract class PushPullCommand<O extends PushPullOptions> extends
             this.total = total;
             this.translated = translated;
             this.approved = approved;
-            translatedPercent = (int) ((translated + approved) * 100 / total);
+            if (total == 0) {
+                // in some case the document has no content, we want to pull the
+                // translation file regardless
+                translatedPercent = 100;
+            } else {
+                translatedPercent = (translated + approved) * 100.0 / total;
+
+            }
         }
 
         public boolean isAboveThreshold(int minimumPercent) {
@@ -357,7 +372,7 @@ public abstract class PushPullCommand<O extends PushPullOptions> extends
             }
         }
 
-        public int getTranslatedPercent() {
+        public double getTranslatedPercent() {
             return translatedPercent;
         }
     }
--- a/zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
+++ b/zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
@@ -314,11 +314,15 @@ public abstract class PushPullCommand<O extends PushPullOptions> extends
         Optional<Map<String, Map<LocaleId, TranslatedPercent>>> optionalStats =
                 Optional.absent();
         O opts = getOpts();
-        if (pullTarget && opts instanceof PullOptions &&
-                ((PullOptions) opts).getMinDocPercent() > 0) {
+        if (needToGetStatistics(pullTarget)) {
             optionalStats = Optional.of(getDocsTranslatedPercent(locales));
         }
         return optionalStats;
+    }
+
+    protected boolean needToGetStatistics(boolean pullTarget) {
+        return pullTarget && getOpts() instanceof PullOptions &&
+                ((PullOptions) getOpts()).getMinDocPercent() > 0;
     }
 
     protected boolean shouldPullThisLocale(
--- a/zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
+++ b/zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
@@ -44,6 +44,7 @@ import org.zanata.client.etag.ETagCacheReaderWriter;
 import org.zanata.client.exceptions.ConfigException;
 import org.zanata.rest.client.RestClientFactory;
 import org.zanata.rest.client.SourceDocResourceClient;
+import org.zanata.rest.client.StatisticsResourceClient;
 import org.zanata.rest.client.TransDocResourceClient;
 import org.zanata.rest.dto.resource.Resource;
 import org.zanata.rest.dto.resource.ResourceMeta;
@@ -67,6 +68,7 @@ public abstract class PushPullCommand<O extends PushPullOptions> extends
     private String modulePrefix;
     protected SourceDocResourceClient sourceDocResourceClient;
     protected TransDocResourceClient transDocResourceClient;
+    protected final StatisticsResourceClient statsClient;
 
     public PushPullCommand(O opts, RestClientFactory clientFactory) {
         super(opts, clientFactory);
@@ -80,6 +82,7 @@ public abstract class PushPullCommand<O extends PushPullOptions> extends
         transDocResourceClient =
                 getClientFactory().getTransDocResourceClient(opts.getProj(),
                         opts.getProjectVersion());
+        statsClient = getClientFactory().getStatisticsClient();
     }
 
     public PushPullCommand(O opts) {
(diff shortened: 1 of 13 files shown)


zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PullCommand.java
```diff
@@ -401,4 +491,26 @@ private void writeTargetDoc(PullStrategy strat, String localDocName,
...
+
+        public TranslatedPercent(long total, long translated, long approved) {
+            this.total = total;
+            this.translated = translated;
+            this.approved = approved;
+            translatedPercent = (int) ((translated + approved) * 100 / total);
+        }
+
+        public boolean isAboveThreshold(int minimumPercent) {
+            if (minimumPercent == 100) {
```
seanf: This should have an explanatory comment.

zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PullOptionsImpl.java
```diff
@@ -166,6 +167,16 @@ public boolean isContinueAfterError() {
     }
 
     @Override
+    public int getMinDocPercent() {
+        return this.minDocPercent;
+    }
+
+    @Option(name = "--min-doc-percent", aliases = "-m", usage = "When specified, will only pull down translation files that have translated or better statistics above the specified percentage")
```
seanf: I'm not sure we should use up one of our single letter options on this.

Plus the help doesn't make it clear that the percentage is integer-only, which may surprise some users.

zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PullCommand.java
```diff
@@ -198,13 +207,19 @@ public void run() throws Exception {
...
-            confirmWithUser("This will overwrite/delete any existing translations in the above directory.\n");
+            confirmWithUser(
+                    "This will overwrite/delete any existing translations in the above directory.\n");
         }
 
         if (getOpts().getPurgeCache()) {
             eTagCache.clear();
         }
 
+        Map<String, Map<LocaleId, TranslatedPercent>> statsMap = null;
```
seanf: Would'n hurt to mark this `@Nullable`.  Also, please add a comment explaining the type params.

zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PullCommand.java
```diff
@@ -330,6 +287,139 @@ public void run() throws Exception {
...
+            // Write the target document
+            writeTargetDoc(strat, localDocName, locMapping,
+                    doc, targetDoc,
+                    transResponse.getHeaders()
+                            .getFirst(HttpHeaders.ETAG));
+        }
+    }
+
+    private boolean shouldPullThisLocale(
+            Map<String, Map<LocaleId, TranslatedPercent>> statsMap,
```
seanf: `@Nullable` would be even better to have here (unless you want to use Optional).

zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PullOptionsImpl.java
seanf: I assume "translated or better" means "Translated or Approved/Reviewed"?  It won't be clear to everyone.  Especially if they just try to read it as a sentence, because "translated" is really the name of a state in the system, but not everyone will know that.  Perhaps recast the sentence entirely, so that we don't have to talk about Translated or Approved at all.

seanf: I think `--min-doc-translated-percent` would be clearer, even if it is longer.  Same goes for the Maven option.

huangp: Do you have a better sentence in mind? I couldn't work out how to convey the meaning of translated and approved state. Also when we implement the feature where project's requiring review will only allow users to download approved state, this sentence will still make sense.

huangp: `--min-doc-translated-percent` is clearer but once we implement the "pull approved only in version requiring review" feature, this will become confusing. Although it's our internal state, but if users don't know the meaning they probably don't know how to use that feature.

seanf: "Only pull documents which are at least PERCENT % translated" along with a
metaVar for PERCENT. Users won't care about the states, they just care
whether it has been translated.  When we implement the feature you mention,
we can update the description to cover Approved, probably along with the
option to override it, etc etc.

I guess if we want the same option also to work for "at least PERCENT %
approved" later on, it might be better to keep the option name generic,
like "--min-doc-percent".

seanf: If you really want to talk in terms of our states, I think you should use
something like Translated and Approved, or "translated" and "approved" (to
make it clear that we are talking specifically about the Zanata states and
their meanings), not just 'translated or better' without quotes.

But I don't think we need to.

huangp: How about "Only pull documents which are at least PERCENT % completed"?

seanf: On 2015-05-19 14:11, Patrick Huang wrote:
> In
> zanata-client-commands/src/main/java/org/zanata/client/commands/pull/PullOptionsImpl.java
> <https://github.com/zanata/zanata-client/pull/63#discussion_r30568152>:
> 
>> @@ -166,6 +167,16 @@ public boolean isContinueAfterError() {
>>      }
>>  
>>      @Override
>> +    public int getMinDocPercent() {
>> +        return this.minDocPercent;
>> +    }
>> +
>> +    @Option(name = "--min-doc-percent", aliases = "-m", usage = "When specified, will only pull down translation files that have translated or better statistics above the specified percentage")
> 
> How about "Only pull documents which are at least PERCENT % completed"?
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zanata/zanata-client/pull/63/files#r30568152>.
> 

Sounds perfect.

-- 
Sean Flanigan

Principal Software Engineer
Localisation Tools Engineering
Red Hat

huangp:
done

seanf:
:+1:

definite:
For File project, it still pull everything even the document does not met the requirement.

definite:
Note that document with 94.98% are still download with --min-doc-percent 95

seanf:
If we handle that correctly, we won't need a special case for 100% by
itself.

aeng:
Sounds like a decimal point calculation issue.

huangp:
I've answered this in irc will answer again here:
We are doing integer division here so the result will be a float point number. In computer it can only represent this with a limit precision. We could certainly use BigDecimal to make the precision big enough plus specific rounding rule to accommodate most cases. But in real world, it's still an approximation. 100% is definitely a special case. 100000000000000000000 total messages with 1 untranslated will still not qualify as 100%. 
Since we don't offer delete old documents (the feature requester clarified this in https://bugzilla.redhat.com/show_bug.cgi?id=1215274#c2), any update to source document will cause a percentage change for existing files. They don't care.
Since the option accept only integer, I chose to only round to integer. Unless someone has specific precision requirement, I tend to keep it simple as it is.

huangp:
@definite last commit will fix file type project.

aeng:
>I've answered this in irc will answer again here:...

:+1:

definite:
I got this 

java.lang.ArithmeticException: / by zero
        at org.zanata.client.commands.PushPullCommand$TranslatedPercent.<init>(PushPullCommand.java:347)
        at org.zanata.client.commands.PushPullCommand.getDocsTranslatedPercent(PushPullCommand.java:278)
        at org.zanata.client.commands.PushPullCommand.prepareStatsIfApplicable(PushPullCommand.java:317)
        at org.zanata.client.commands.pull.RawPullCommand.run(RawPullCommand.java:149)
        at org.zanata.client.commands.ConfigurableCommand.runWithActions(ConfigurableCommand.java:110)
        at org.zanata.client.commands.ArgsUtil.runCommand(ArgsUtil.java:48)
        at org.zanata.client.ZanataClient.processArgs(ZanataClient.java:170)
        at org.zanata.client.ZanataClient.main(ZanataClient.java:95)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:497)
        at org.codehaus.mojo.exec.ExecJavaMojo$1.run(ExecJavaMojo.java:297)
        at java.lang.Thread.run(Thread.java:745)


When I ran this:

zanata-cli -e -B pull --min-doc-percent 90 -t /tmp/spacewalk100 --disable-ssl-cert --url https://fedora.zanata.org/ --project-version frontend

zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
```diff
@@ -245,4 +257,108 @@ protected void storeETagCache() {
...
+                            .get(localDocName).get(serverLocale)
+                    .getTranslatedPercent());
+        }
+        return !optionalStats.isPresent()
+                || optionalStats.get().get(localDocName).get(serverLocale)
+                        .isAboveThreshold(minDocPercent);
+    }
+
+    protected static class TranslatedPercent {
+        protected final int translatedPercent;
```
seanf: Why protected?

If you change int to double, it should be possible to get far better precision in the percentage checks.

zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
```diff
@@ -245,4 +257,108 @@ protected void storeETagCache() {
...
+        private final long total;
+        private final long translated;
+        private final long approved;
+
+        public TranslatedPercent(long total, long translated, long approved) {
+            this.total = total;
+            this.translated = translated;
+            this.approved = approved;
+            translatedPercent = (int) ((translated + approved) * 100 / total);
```
seanf: Try this:

    translatedPercent = (translated + approved) * 100.0 / total;

zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
```diff
@@ -245,4 +257,108 @@ protected void storeETagCache() {
...
+            // if minimum percent is 100, we will compare exact number so that
+            // rounding issue won't affect the result
+            if (minimumPercent == 100) {
+                return total == translated + approved;
+            } else {
+                return translatedPercent >= minimumPercent;
+            }
+        }
+
+        public int getTranslatedPercent() {
```
seanf: int -> double

seanf:
On 2015-05-22 14:01, Patrick Huang wrote:
> I've answered this in irc will answer again here:
> We are doing integer division here so the result will be a float point
> number.

What?  Don't you mean the result will be an int or long?  (You can of
course force an operand to be a double if you want a double answer.)

> In computer it can only represent this with a limit precision.
> We could certainly use BigDecimal to make the precision big enough plus
> specific rounding rule to accommodate most cases. But in real world,
> it's still an approximation. 100% is definitely a special case.
> 100000000000000000000 total messages with 1 untranslated will still not
> qualify as 100%.

No more a special case than the one @definite found.
100000000000000000000 total with 50000000000000000001 untranslated is
not quite 50%.

But I'm not worried about cases with that many messages, or even
(Long.MAX_VALUE - 100) translated vs (Long.MAX_VALUE) total.  The
int-based calculations start going wrong at 101 total messages.

> Since we don't offer delete old documents (the feature requester
> clarified this in
> https://bugzilla.redhat.com/show_bug.cgi?id=1215274#c2), any update to
> source document will cause a percentage change for existing files. They
> don't care.
> Since the option accept only integer, I chose to only round to integer.
> Unless someone has specific precision requirement, I tend to keep it
> simple as it is.

It's okay to restrict the user option --min-doc-percent to an integer
(in the absence of a user request for "85.1%" support), but that doesn't
mean we should use low precision calculations.

If the user says "minimum percentage" is 95%, it is incorrect to return
a document where the percentage is anything less than 95%.  Just like it
is incorrect for 100%.

Why don't you just perform the calculations in `TranslatedPercent` using
double precision?  Should be a three line fix.


You would need to use a smarter (non floating point) implementation if
you want to solve the 100000000000000000000 problem properly (including
the 85% case), but I would settle for the double precision solution.
(But keep the 100% special case too, if you like.)


-- 
Sean Flanigan

Principal Software Engineer
Localisation Tools Engineering
Red Hat

zanata-client-commands/src/main/java/org/zanata/client/commands/PushPullCommand.java
huangp: protected is a side effect from IDE refactoring. Will fix.

huangp:
I know double will have better precision than int and I know I can force an operand to be double to get double (I used to work in a bank and this type of calculation is not uncommon).
My point is, I don't think people will care if the precision is off by 0.5% or lower unless in 100% case. "100000000000000000000 total with 50000000000000000001 untranslated is not quite 50%" (if I were a project maintainer and I specify 50 as min I don't care the system give my 49.5%). If you guys are so concern about the correctness after decimal point I will make it more precise.

seanf:
I care.  Please do.  Fixing it with double will be easier than explaining
to users why we we didn't do (exactly) what they asked.

seanf:
@huangp Looks like there are some merge conflicts.

huangp:
@seanf added the INFO log in last commit

seanf:
:+1:

definite:
When using:
export MIN_PERCENT=92
zanata-cli -e -B pull --min-doc-percent $MIN_PERCENT -t /tmp/ibus-chewing$MIN_PERCENT --url https://fedora.zanata.org/    

As https://fedora.zanata.org/iteration/view/ibus-chewing/master/languages/fr
French (fr) is 91.42%, but still get pulled.

huangp:
91.42% is word based stats. The message based stats for that project version and locale is 92%.

definite:
Test cases:
 - [x] File project
   - [x] 100%
   - [x] Celling
   - [x] floor
   - [x] 0%
 - [x] Gettext project
    - [x] 100%
   - [x] Celling
   - [x] floor
   - [x] 0%

Note that xliff bug is addressed by zanata/zanata-common#12.
https://bugzilla.redhat.com/show_bug.cgi?id=1225290

:+1:

aeng:
Wrong branch to merged into. Integration/master is missing this PR.

PR #64, opened 2015-05-20
rhbz1222730 - add options to use cache in java client

huangp:


seanf:
I recommend that @carlosmunoz review this one.  I assume these options are similar to the Maven options?

huangp:
Yes. except maven has different naming convention and the way to set it.

seanf:
Sure, that's to be expected.

carlosmunoz:
So, is this all that was missing? Looks like the variables were already there, and the code to use them as well, we just need to be able to set the vars. If that's the case, then :+1:

definite:
:+1:

PR #68, opened 2015-06-21
Add doc for 0install

aeng:
0install documentation:
https://bugzilla.redhat.com/show_bug.cgi?id=1229554


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-client/pull/68" target="_blank">ReviewNinja

docs/0install/setup.md
```diff
@@ -0,0 +1,28 @@
+The 0install distribution of the client is to allow zanata-cli to be available on multiple platform. This distribution requires [0install](http://0install.net/) to run.
+
+## Prerequisites
+
+1. Install [0install](http://0install.net/injector.html) on your machine.
```
seanf: This should probably be clarified for Mac, because the OSX page is dominated by "Source Installation".  I think @lukebrooker had to use the pre-compiled package bundle, which is easy to miss amongst all the information about building from source.

docs/installation.md
```diff
@@ -15,6 +15,10 @@ sudo yum install zanata-client
 
 See [Ivy installation](/ivy/installation) for more information.
 
+## Installation with 0install
+
+See [Ivy installation](/0install/setup) for more information.
```
seanf: This still says "Ivy"...

docs/0install/setup.md
aeng: Cofirmed that with the jre dependency in place, Mac won't be able to run even if it has local valid JRE installed. So for now, I've commented out the dependency until they fix it in the 0install repo. That means the JRE installation will be applying to all OS for now.

seanf: > That means the JRE installation will be applying to all OS for now.

Meaning all users will have to install a JRE manually?  If so, it might be
better to make a separate feed for Macs.

aeng: The advantage of having single feeds file is when the dependency is fix, mac user just need to do  0install update zanata-cli. If separated,  we will need to force them to change url again.

seanf: Right, but the disadvantage is that we are inconveniencing all users
because of a Mac problem, when the aim of this work is to make things
simpler for users.

Anyway, didn't Luke succeed in using the feed even when it depended on
http://repo.roscidus.com/java/openjdk-jre, by installing the JDK manually?

aeng: Luke successfully install without the dependency.

djansen-redhat:
:+1: Looks good

PR #72, opened 2015-07-21
fix(ux,doc): update doc regarding file mapping rule

huangp:
plus log warning where there is no mapping rule can be applied


definite:
Please also back port to release.

huangp:
backported to release at be75556a1

PR #79, opened 2015-09-16
feat(ZNTA-664): disable copytrans by default

seanf:


carlosmunoz:
:+1: 
How long do you think we should leave the warning up?

seanf:
Until we get a lot of complaints about the warning, or we remove copyTrans
from Zanata entirely. ;-)

On 16 September 2015 at 19:14, Carlos Munoz <notifications@github.com>
wrote:

> [image: :+1:]
> How long do you think we should leave the warning up?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zanata/zanata-client/pull/79#issuecomment-140681253>.
>



-- 
Sean Flanigan

Principal Software Engineer
Globalisation Tools Engineering
Red Hat

seanf:
@carlosmunoz could you please have a look at the last three commits?

PR #80, opened 2015-10-22
fix build after jersey version upgrade to 1.19

huangp:
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-client/80)


seanf:
@huangp We probably don't need this any more, in view of https://zanata.atlassian.net/browse/ZNTA-1271

huangp:
I don't remember what this is for. Since the build is green now I don't think this is needed.

PR #81, opened 2015-10-23
Stop api from forcing jersey upgrade

seanf:
It's not ideal, but this should fix the compilation error caused by https://github.com/zanata/zanata-api/commit/d8948c917c96b49b6dd135878ae07c07868dcb30.

Note that Fedora 21 is still on Jersey 1.17.1.


huangp:
:+1:

PR #83, opened 2015-11-03
Update OSX install docs with Homebrew [skip ci]

Fryguy:
This depends on my PR to homebrew to add the zanata-client formula: https://github.com/Homebrew/homebrew/pull/44515

Once that is in, this PR updates the documentation to use that.  I'm not sure this is the same as what's on the website, so if there's another place I need to update this documentation, please let me know.

Incidentally, when I go to http://zanata.org/help/cli-install/, I no longer see the OSX installation page.  I was pretty sure it used to be there, but I don't see it anymore...did something change?

### DIFF
--- a/docs/installation/osx-installation.md
+++ b/docs/installation/osx-installation.md
@@ -1,6 +1,15 @@
+## With Homebrew
+
+1. Install [Homebrew](http://brew.sh/): `ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
+  - If you already have Homebrew installed, run `brew update` to pull in the latest formulae.
+2. Install the zanata-client: `brew install zanata-client`
+3. Now you can run `zanata-cli --help` for more options.
+
+## With 0Install
+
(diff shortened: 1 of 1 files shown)


Fryguy:
Oh I remember now, I found it here: http://zanata-client.readthedocs.org/en/latest/installation/osx-installation/  .  Not sure why those docs are more thorough than the ones found on zanata.org .

lukebrooker:
@Fryguy thanks for this. I've been wanting to add Zanata client to Homebrew but haven't had a chance to work out how. We'll merge this as soon as it's in Homebrew.

You are correct on the documentation. The website help pages should be redirecting to the new docs pages. I'll check on the status of the redirects.

lukebrooker:
Seems like it is published now. Thanks again @Fryguy. I've tested it out and it works great.

I'll add one step to the docs, asking to install Java using `brew cask install java` if you don't have it, as homebrew doesn't do that for you.

BTW, do I just submit another pull request to update the version on homebrew? We actually realeased a new version of the client yesterday.

Fryguy:
@lukebrooker Thanks!  Yes, if you want to bump the version on homebrew, just make a new PR there and update the url and sha256 lines.

PR #85, opened 2015-11-09
Fix warning "Use of '_' might not be supported..."

seanf:
Without this change, the source may not compile under future versions of Java.


huangp:
:+1:

seanf:
@djansen-redhat @definite Do you need to check this?

definite:
See also: https://zanata.atlassian.net/browse/ZNTA-519

PR #87, opened 2015-12-08
docs(install): add java to client yum install instructions

davidmason:
Translators are having issues with both 0install and yum installation on RHEL7. Adding this got one of them working, so adding it to the docs for now.

Targeting `master` on purpose because the jenkins system is not operating and would not merge the changes in a timely manner for translators to use them.


davidmason:
@djansen-redhat please review this, just to be extra-cautious before merging directly to master.

djansen-redhat:
:+1:

PR #88, opened 2015-12-10
docs(0install): Update 0install installation guide

aeng:
https://zanata.atlassian.net/browse/ZNTA-843


aeng:
:+1:  from @definite

PR #89, opened 2015-12-16
Update windows-installation.md

nitech:
Seems I need to simplify the 0install command to the above. Now install works (although I get the error "The specified executable is not a valid application for this OS platform" when trying to start the app)


nitech:
Quick as a lightning bolt. Yes please.

aeng:
Please target this change to our integration/master branch.

PR #90, opened 2015-12-17
Update windows-installation.md

nitech:
Seems I need to simplify the 0install command to the above. Now install works (although I get the error "The specified executable is not a valid application for this OS platform" when trying to start the app). Ref #89 

Did the pull request end up on the correct branch now?

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-client/90)


nitech:
By the way @aeng, I want to report the issue mentioned in this pull request, but I can't seem to find the "new issue" button anywhere. Can you point me in the right direction?

aeng:
@nitech , you can file the issue in zanata.atlassian.net

djansen-redhat:
I tested this on Windows 10.
The 0install process still requires the "add" command, and once installed "zanata-cli --help" did not work (bad command or file name).

definite:
@djansen-redhat I suppose that's because it is not in PATH.

aeng:
one of the issue with 0install in windows was supporting the  .sh file which what its doing at the moment. We might need to modify the client to support executable file in windows.

aeng:
We have identified the issue in Windows platform that will prevent execution of zanata-cli, missing of executable file. We have updated the windows installation guide (http://zanata-client.readthedocs.io/en/latest/installation/windows-installation/) to use jar file until we implement 0install support for zanata-cli.

PR #95, opened 2016-02-22
Update manual-installation.md to fix symlinking

pixelrust:
Added `--relative` option to instructions for symbolic linking. 
`ln` needs this option to create working symbolic links when a non-absolute target path is provided, as in the current examples.

PR #96, opened 2016-02-23
Update manual-installation.md to fix symlinking

pixelrust:
Added `--relative` option to instructions for symbolic linking.
`ln` needs this option to create working symbolic links when a non-absolute target path is provided, as in the current examples.

Same change as in #95, but to a different branch as per @aeng 's request.


carlosmunoz:
@definite assigning this for a quick review

PR #97, opened 2016-02-23
Contributing

definite:
It is same with CONTRIBUTING.md in zanata-server except the "Test against wildfly"


davidmason:
It looks like there are a bunch of other changes in here that aren't related to `CONTRIBUTING.md`. What are those about?

definite:
Right, it should target against release instead of master, closing this one.

definite:
New PR is #99

PR #101, opened 2016-03-01
ZNTA-930 - zanata init can handle file type project

huangp:


huangp:
@zanata-jenkins retest this please

zanata-client-commands/src/main/java/org/zanata/client/commands/init/SourceConfigPrompt.java
```diff
@@ -117,6 +125,24 @@ SourceConfigPrompt promptUser() throws Exception {
...
+            RestClientFactory clientFactory = getClientFactory(pushOptions);
+            FileResourceClient client =
+                    clientFactory.getFileResourceClient();
+            rawDocumentTypes = client.acceptedFileTypes();
+            // this answer is not persisted in zanata.xml so user will still need to type it when they do the actual push
+            console.printfln(Hint, PushOptionsImpl.fileTypeHelp);
+            console.printf(Question, get("file.type.prompt"));
+            String answer = console.expectAnyNotBlankAnswer();
+            ((PushOptionsImpl) pushOptions).setFileTypes(answer);
+        }
```
davidmason: This file is so big, I think we could split it into different files that each handle querying the user for one or a few related things. Not necessary right now, but we should consider it if we add anything else. It would make it much easier to see (and change) the overall flow of prompts.

zanata-client-commands/src/main/java/org/zanata/client/commands/init/SourceConfigPrompt.java
```diff
@@ -191,4 +218,92 @@ public String getExcludes() {
...
+    private SrcDocsFinder makeSrcDocFinder() {
+        String projectType = pushOptions.getProjectType();
+        if (projectType.equalsIgnoreCase(File.name())) {
+            return new RawSrcDocsFinder(pushOptions);
+        } else {
+            return new OtherSrcDocsFinder(pushOptions);
+        }
+    }
+
+    interface SrcDocsFinder {
```
davidmason: Is this repeating the logic in the actual push/pull commands?
This and its implementations are big enough that I would prefer them in their own file. If possible it would be even better if these can be re-used both here and in the other commands. If the code is in 2 different places, there is a risk that the files found in the init command will not match the files found in push/pull.

davidmason: Ah, I see these are really just wrappers that call to the push/pull strategy, to there is not the reuse problem. We can probably structure the whole client code to make reuse of pieces of the push/pull algorithm easier, but not necessary to do that now.

zanata-client-commands/src/main/java/org/zanata/client/commands/init/TransConfigPrompt.java
```diff
@@ -55,7 +59,8 @@
     private final ConfigurableProjectOptions opts;
     private final Set<String> srcFilesSample;
     private final PullOptionsImpl pullOptions;
-    private final PullCommand pullCommand;
+    private final TransFilePathFinder
+            transFilePathFinder;
```
davidmason: This line doesn't look long enough to wrap.

zanata-client-commands/src/main/java/org/zanata/client/commands/init/TransConfigPrompt.java
```diff
@@ -123,19 +132,61 @@ TransConfigPrompt promptUser() throws Exception {
...
+            Optional<String> translationFileExtension =
+                    Optional.fromNullable(targetFileExt);
+            File file = transFileResolver.resolveTransFile(
+                    QualifiedSrcDocName.from(srcDoc),
+                    localeMapping, translationFileExtension);
+            return file.getPath();
+        }
+    }
+
+    static class OtherTransFilePathFinder implements TransFilePathFinder {
```
davidmason: If it is specifically not raw files, I suggest `NonRawTransFilePathFinder`, since "Other" is not meaningful on its own.

zanata-client-commands/src/main/java/org/zanata/client/commands/init/SourceConfigPrompt.java
```diff
@@ -117,6 +125,24 @@ SourceConfigPrompt promptUser() throws Exception {
...
+        if (File.name().equalsIgnoreCase(pushOptions.getProjectType())) {
+
+            console.blankLine();
+            console.printfln(Question, get("project.file.type.question"));
+            RestClientFactory clientFactory = getClientFactory(pushOptions);
+            FileResourceClient client =
+                    clientFactory.getFileResourceClient();
+            rawDocumentTypes = client.acceptedFileTypes();
+            // this answer is not persisted in zanata.xml so user will still need to type it when they do the actual push
+            console.printfln(Hint, PushOptionsImpl.fileTypeHelp);
```
davidmason: Is this actually showing the user the accepted file types fetched from client in line 138?

zanata-client-commands/src/main/resources/prompts.properties
```diff
@@ -43,6 +43,8 @@ includes.prompt=Includes (leave blank if not applicable):
 excludes.question=Do you want to define any Excludes?
 excludes.usage= - Wildcard pattern to exclude files and directories. Defined it as "configuration.properties,build.properties".
 excludes.prompt=Excludes (leave blank if not applicable):
+project.file.type.question=What file types do you want to use?
+file.type.prompt=Please enter file types in comma separated format (e.g. XML_DOCUMENT_TYPE_DEFINITION,IDML[txt]):
```
davidmason: Suggest changing `IDML[txt]` to `PLAIN_TEXT[md;markdown]`, because no one would usually name their IDML files with .txt extension, but people do have to use plain text for their markdown files at the moment.

zanata-client-commands/src/test/java/org/zanata/client/commands/init/SourceConfigPromptTest.java
```diff
@@ -109,4 +109,45 @@ public void allowUserToRetry() throws Exception {
...
+
+    @Test
+    public void canHandleFileProjectType() throws Exception {
+        when(fileClient.acceptedFileTypes()).thenReturn(ImmutableList.copyOf(DocumentType.values()));
+        // here we use absolute path because we create temp files in there
+        String expectedSrcDir = tempFolder.getRoot().getAbsolutePath() + "/resources";
+        ConsoleInteractor console =
+                MockConsoleInteractor.predefineAnswers(
+                        expectedSrcDir, "messages.txt",
+                        "*Excluded.txt", "PLAIN_TEXT[txt]", "y");
```
davidmason: Default extension for `PLAIN_TEXT` is already `[txt]`, so it would be better to use a different combination that is not default to prove that it is using both components.

zanata-client-commands/src/test/java/org/zanata/client/commands/init/TransConfigPromptTest.java
```diff
@@ -109,4 +109,29 @@ public void allowUserToRetry() throws Exception {
...
+
+        prompt = prompt.promptUser();
+
+        assertThat(opts.getTransDir(), Matchers.equalTo(new File(".")));
+        List<String> capturedPrompts =
+                MockConsoleInteractor.getCapturedPrompts(console);
+        assertThat(capturedPrompts, Matchers.hasItems(
+                "        ./de/a.txt",
+                "        ./de/b.txt"));
+    }
```
davidmason: Do we have any tests that look at what the user would actually see? It would be great to have tests for the console interactions something like:

```
userReads("What file types do you want to use?");
userReads("Hint: ...");
userReads("Please enter file types in comma separated format (e.g. XML_DOCUMENT_TYPE_DEFINITION,IDML[txt]):");
userTypes("PLAIN_TEXT");
userReads(...);
```

so it is easy to see what the real interaction should be like, and easy to change it with TDD.

davidmason:
@huangp finished review. Suggested a few minor changes, and a few ideas for bigger improvements, but nothing that should block merge.

zanata-client-commands/src/main/java/org/zanata/client/commands/init/SourceConfigPrompt.java
huangp: No... initially I was planning to show a list and let user choose. But then I realized it's a multiple choice question (plus optional extension) and it's too hard to implement. So now I just reuse the help message not the actual result fetched from server. I will remove this.

huangp: It is used in org.zanata.client.commands.init.SourceConfigPrompt.RawSrcDocsFinder#filteredFileExtensions down below. I will move it to there.

zanata-client-commands/src/test/java/org/zanata/client/commands/init/SourceConfigPromptTest.java
```diff
@@ -109,4 +109,45 @@ public void allowUserToRetry() throws Exception {
...
+
+    @Test
+    public void canHandleFileProjectType() throws Exception {
+        when(fileClient.acceptedFileTypes()).thenReturn(ImmutableList.copyOf(DocumentType.values()));
+        // here we use absolute path because we create temp files in there
+        String expectedSrcDir = tempFolder.getRoot().getAbsolutePath() + "/resources";
+        ConsoleInteractor console =
+                MockConsoleInteractor.predefineAnswers(
+                        expectedSrcDir, "messages.md",
+                        "*Excluded.txt", "PLAIN_TEXT[md,txt]", "y");
```
davidmason: Does this actually work with `...[md,txt]`? According to docs it should be semicolon separator for tile extensions: `...[md;txt]`

huangp: oh this test won't test that. It only test the input matches the option. I've tested it locally and it only works with ;

davidmason:
:+1:

definite:
:+1:

PR #106, opened 2016-03-22
Simplify parent

seanf:


pom.xml
```diff
@@ -212,6 +214,30 @@
         </exclusions>
       </dependency>
       <dependency>
+        <groupId>org.powermock</groupId>
```
aeng: So we using powermock now.... nice

aeng:
:+1:

pom.xml
seanf: We already were, but this configuration was previously inherited from parent.

PR #112, opened 2016-04-14
Rename QualifiedSrcDocName and UnqualifiedSrcDocName

seanf:
Qualified doc name already has a meaning in the client: a document name
with module prefix, if any. Unqualified doc name is the document within
a module. (If Maven multi-modules aren't used, qualified and
unqualified doc name are the same.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,12 @@ cache:
 
 language: java
 
+before_install:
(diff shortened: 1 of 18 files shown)


aeng:
:+1:

seanf:
@djansen-redhat Anything from you? It's mostly a straight refactoring of classnames, plus a few  error message tweaks.

PR #119, opened 2016-06-16
feat(pom): add fedora profile

definite:
https://zanata.atlassian.net/browse/ZNTA-1180

This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-client/119)


seanf:
@definite, instead of changing plugin configuration, did you try just setting the property required.maven?

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/119#-:-KKQh4njFFft7yCrNC-7:1204942688)*

seanf:
I mean: did you try setting the system property *inside the new fedora profile*, I mean? The idea of the profile is good.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[pom.xml, line 267 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/119#-KKQhfNZ4G5x1MG24siH:-KKQhfNZ4G5x1MG24siI:-999208101) ([raw file](https://github.com/zanata/zanata-client/blob/10323fa3199cb1ce97679f0e1ea89972dc7dd617/pom.xml#L267)):*
> ```XML
>   <profiles>
>     <profile>
>       <id>fedora</id>
> ```

You should add property activation, so that the profile can be activated with a system property. Using -P to activate profiles doesn't mix well with the use of other profiles.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/119#-:-KKQhRcqE6GjOE2xUGSP:1324094105)*

definite:
The -Drequired.maven=3.3.3 does work, thus closing this PR

seanf:
@definite I think it might still make sense to have a fedora profile, as long as it's activated by a system property, and (in this case) just sets the property `required.maven`.

PR #126, opened 2016-08-04
ZNTA-1271 - migrate JAX-RS client back to resteasy

huangp:
https://zanata.atlassian.net/browse/ZNTA-1271

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-client/126)


seanf:
Reviewed 38 of 38 files at r1.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[zanata-cli/src/test/java/org/zanata/client/ClientToServerTest.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2E2GmBXPtaFVupMl3:-KR2E2GmBXPtaFVupMl4:b-2q5lza) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-cli/src/test/java/org/zanata/client/ClientToServerTest.java#L78)):*
> ```Java
>             assertThat("Client will display meaningful message for 503",
>                     throwableCapture.getValue().getMessage()
>                             .contains("Service is currently unavailable. Please check outage notification or try again later."));
> ```

I don't think we should be so specific, since it can make the test fragile. Are we able to check for the numeric 503 status by any chance? (Perhaps by improving ResponseStatusFilter?) Otherwise, how about `containsIgnore("unavailable")`?

---

*[zanata-client-commands/src/main/java/org/zanata/client/commands/init/ProjectPrompt.java, line 205 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2Ef3TVL_Je5oC6Fbj:-KR2Ef3TVL_Je5oC6Fbk:bl0emgn) ([raw file](https://github.com/zanata/zanata-client/blob/d9b4b5e356b08e3b8bca721ce573761c6316a07f/zanata-client-commands/src/main/java/org/zanata/client/commands/init/ProjectPrompt.java#L205)):*
> ```Java
>             projectClient.put(project);
>         } catch (UniformInterfaceException e) {
>             if (e.getResponse().getStatus() >= 399) {
> ```

Is there really no way we can get the status code?

---

*[zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java, line 463 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2GDvvFgfJkGmp_r9q:-KR2GDvw5wrSHonk5f8U:b-x8r5fb) ([raw file](https://github.com/zanata/zanata-client/blob/d9b4b5e356b08e3b8bca721ce573761c6316a07f/zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java#L463)):*
> ```Java
>                     throw new RuntimeException(String.format(
>                             "%n * Error Message: %s;%n * Response From Server: %s]",
>                             e.getMessage(), entity));
> ```

There must be some way of preserving the HTTP status and entity which were in the ClientResponse before the exception was thrown.

---

*[zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java, line 780 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2F0yxAyMJP6s3Y_yb:-KR2F0yxAyMJP6s3Y_yc:bybf5y5) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java#L780)):*
> ```Java
>         } catch (Exception failure) {
>             // 404 - Probably because of an old server
>             if (failure instanceof NotFoundException) {
> ```

Couldn't you just `catch (NotFoundException)`?

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java, line 59 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2IgRFz_1vwusuayR3:-KR2IgRGjbbqjrlj2yx5:b-m8siqw) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java#L59)):*
> ```Java
>     static {
>         // we need to pass the MultipartForm annotations when making the entity
>         // otherwise Resteasy can't find the writer for multipart form
> ```

Um, what? Is this explained somewhere?

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 54 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HE5DSvaejhzyN95O:-KR2HE5DSvaejhzyN95P:btnx78u) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L54)):*
> ```Java
>                 Entity.entity(glossaryEntries, MediaType.APPLICATION_JSON_TYPE);
> 
>         webResource().path("entries").request(MediaType.APPLICATION_JSON_TYPE)
> ```

We're switching from XML to JSON. Why here, specifically?

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HN5L3GSShnPvoJEj:-KR2HN5L3GSShnPvoJEk:bb8coiu) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L78)):*
> ```Java
> 
>     public void deleteAll() {
>         webResource().request(MediaType.APPLICATION_XML_TYPE).delete();
> ```

Multiple instances in multiple files: Instead of repeating `request(MediaType.APPLICATION_XML_TYPE)` everywhere, can't we set this up in a central place somewhere?

That's a lot of extra places to update if we ever want to switch from XML to JSON.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/InvalidContentTypeFilter.java, line 77 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2Hwj3in8i6tCoA3MJ:-KR2Hwj4paNmZ5a8-mwL:b-jhi79u) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/InvalidContentTypeFilter.java#L77)):*
> ```Java
>             log.error(ERROR_MSG);
>             String snippet = String.format(
>                     "Wrong content type received: [%s]",
> ```

Where's the actual snippet? What happened to extracting the page title from the HTML response?

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/RedirectFilter.java, line 49 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2JIP8djWMCXc2T5ON:-KR2JIP8djWMCXc2T5OO:b-6qhblp) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/RedirectFilter.java#L49)):*
> ```Java
>                 Response.Status.Family.REDIRECTION) {
>             throw new IllegalStateException(
>                     "Received status " + responseContext.getStatusInfo() +
> ```

Does this error message at least tell the user what the new URL is?

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 100 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2LfIy2UTWwXDfJ0df:-KR2LfIzrmjOH7LtXFzx:b-pyfg0h) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L100)):*
> ```Java
>                 throw new NotFoundException(msg);
>             }
>             throw new RuntimeException(msg);
> ```

It might be a good idea to have this class throw an exception which can hold on to the HTTP status code, instead of RuntimeException. Or one of several exceptions, depending  on the problem.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

huangp:
Review status: 32 of 38 files reviewed at latest revision, 10 unresolved discussions.

---

*[zanata-cli/src/test/java/org/zanata/client/ClientToServerTest.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2E2GmBXPtaFVupMl3:-KR2OEwcMObOA_aRfYBC:b-896fix) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-cli/src/test/java/org/zanata/client/ClientToServerTest.java#L78)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> I don't think we should be so specific, since it can make the test fragile. Are we able to check for the numeric 503 status by any chance? (Perhaps by improving ResponseStatusFilter?) Otherwise, how about `containsIgnore("unavailable")`?
</details>
Done.

---

*[zanata-client-commands/src/main/java/org/zanata/client/commands/init/ProjectPrompt.java, line 205 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2Ef3TVL_Je5oC6Fbj:-KR2OS5jjmdVHojW671e:b-tip882) ([raw file](https://github.com/zanata/zanata-client/blob/d9b4b5e356b08e3b8bca721ce573761c6316a07f/zanata-client-commands/src/main/java/org/zanata/client/commands/init/ProjectPrompt.java#L205)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Is there really no way we can get the status code?
</details>
done. See changes made in ResponseStatusFilter

---

*[zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java, line 463 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2GDvvFgfJkGmp_r9q:-KR69DhnairKw3qZofrl:b55y0kv) ([raw file](https://github.com/zanata/zanata-client/blob/d9b4b5e356b08e3b8bca721ce573761c6316a07f/zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java#L463)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> There must be some way of preserving the HTTP status and entity which were in the ClientResponse before the exception was thrown.
</details>
Now the message will contain status code and entity if it's from ResponseStatusFilter

---

*[zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java, line 780 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2F0yxAyMJP6s3Y_yb:-KR2PRXr645GQ_zFtIQV:b-896fix) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-client-commands/src/main/java/org/zanata/client/commands/push/PushCommand.java#L780)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Couldn't you just `catch (NotFoundException)`?
</details>
Done.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java, line 59 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2IgRFz_1vwusuayR3:-KR2PpZtv-Oxvqr5SmaD:bb664vb) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java#L59)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Um, what? Is this explained somewhere?
</details>
When creating the entity, down at line 112, we need to pass the annotation for it to recognize the form being a multipart form. So that it can find the appropriate writer. Not sure if this is a defect or missing feature from resteasy. Maybe this is designed to be used by their proxy solution.  

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 54 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HE5DSvaejhzyN95O:-KR2Qj4D2IHBlmaPH0Gs:b-5cpunj) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L54)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> We're switching from XML to JSON. Why here, specifically?
</details>
Because the entity is a List. To marshal it to xml (with current resteasy library), looks like it requires another container type (extends List) and is annotated with XmlRootElement. Using json will work out of the box.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HN5L3GSShnPvoJEj:-KR2Qzj8hgDhaPe6LjvS:brtujo6) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L78)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Multiple instances in multiple files: Instead of repeating `request(MediaType.APPLICATION_XML_TYPE)` everywhere, can't we set this up in a central place somewhere?
> 
> That's a lot of extra places to update if we ever want to switch from XML to JSON.
</details>
Not everything uses the same type. Some are using zanata specific media type, some use octet stream, some use json.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/InvalidContentTypeFilter.java, line 77 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2Hwj3in8i6tCoA3MJ:-KR2RabvCQ5GeotwwwOb:bywrrkk) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/InvalidContentTypeFilter.java#L77)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Where's the actual snippet? What happened to extracting the page title from the HTML response?
</details>
From memory it was tricky to read the entity out as string/text. So I didn't bother extracting the page title again.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/RedirectFilter.java, line 49 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2JIP8djWMCXc2T5ON:-KR2SRGxlw9zlgcoDihU:b-896fix) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/RedirectFilter.java#L49)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Does this error message at least tell the user what the new URL is?
</details>
Done.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 100 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2LfIy2UTWwXDfJ0df:-KR69Pd8oTfZB71M2PdG:b-896fix) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L100)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> It might be a good idea to have this class throw an exception which can hold on to the HTTP status code, instead of RuntimeException. Or one of several exceptions, depending  on the problem.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

seanf:
Reviewed 6 of 6 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java, line 59 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2IgRFz_1vwusuayR3:-KR6Pzk9OASMCgyC6fYL:b2ve8hj) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java#L59)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> When creating the entity, down at line 112, we need to pass the annotation for it to recognize the form being a multipart form. So that it can find the appropriate writer. Not sure if this is a defect or missing feature from resteasy. Maybe this is designed to be used by their proxy solution.  
</details>
But where did the idea for this minor monstrosity come from? An answer on stackoverflow? A blog post? Or did it come to you in a nightmare?

I still don't understand why the method we are calling is designed to receive Annotation objects.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 54 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HE5DSvaejhzyN95O:-KR6PajELiofNAdztRnf:b-dp7bks) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L54)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> Because the entity is a List. To marshal it to xml (with current resteasy library), looks like it requires another container type (extends List) and is annotated with XmlRootElement. Using json will work out of the box.
</details>
That sounds like a kludge. It might be better to switch completely to JSON, instead of switching to a mixture.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HN5L3GSShnPvoJEj:-KR6PBc2V-Sv-bZ71ncV:bwbhkll) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L78)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> Not everything uses the same type. Some are using zanata specific media type, some use octet stream, some use json.
</details>
Right, but can't we set the default to MediaType.APPLICATION_XML_TYPE and then override it for the few cases which use something else?

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 53 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR6N6yJ4h9aqEk8-2Z2:-KR6N6yJ4h9aqEk8-2Z3:bu4tcl4) ([raw file](https://github.com/zanata/zanata-client/blob/67153ea411b52c6c0e8c9b025d1a64621ae9e618/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L53)):*
> ```Java
> 
>         if (responseStatus == Response.Status.SERVICE_UNAVAILABLE) {
>             throw new RuntimeException("Service is currently unavailable. " +
> ```

Perhaps this should be ServerErrorException, as below?

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 70 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR6NUSYTriP2KhIBCNz:-KR6NUSYTriP2KhIBCO-:bm4cp7s) ([raw file](https://github.com/zanata/zanata-client/blob/67153ea411b52c6c0e8c9b025d1a64621ae9e618/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L70)):*
> ```Java
>                         "Server returned a redirect. You must change your url option or config file.";
>             }
>             throw new RuntimeException(message);
> ```

It might be better to create a custom exception here.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 78 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR6NZb9DbwJqnVgBcZh:-KR6NZb9DbwJqnVgBcZi:b-7mth5q) ([raw file](https://github.com/zanata/zanata-client/blob/67153ea411b52c6c0e8c9b025d1a64621ae9e618/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L78)):*
> ```Java
>                             entity);
>             if (responseStatus == Response.Status.UNAUTHORIZED) {
>                 throw new RuntimeException("Incorrect username/password");
> ```

It might be better to create a custom exception here, unless there's one in JAX-RS we could use.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 91 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR6OC1-H41HQPbpfVRl:-KR6OC10f7heHYqjuaeK:bio4xe4) ([raw file](https://github.com/zanata/zanata-client/blob/67153ea411b52c6c0e8c9b025d1a64621ae9e618/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L91)):*
> ```Java
>     private static boolean isRedirect(int statusCode) {
>         return statusCode == Response.Status.MOVED_PERMANENTLY.getStatusCode()
>                 || statusCode == 302;
> ```

You could use `Response.Status.FOUND.getStatusCode()` or just use the numbers 301 and 302, but I think it would be best to pick constants or numbers, not one of each.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

huangp:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java, line 59 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2IgRFz_1vwusuayR3:-KR6ZTCirSGmge3EHH3I:b-k6zwud) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java#L59)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> But where did the idea for this minor monstrosity come from? An answer on stackoverflow? A blog post? Or did it come to you in a nightmare?
> 
> I still don't understand why the method we are calling is designed to receive Annotation objects.
</details>
I actually trace it down to its source code. The registered writer (in a map) need to match both the media type and the annotation.  We don't have a class like AnnotationLiteral in client so I came up with this idea to get hold of that specific annotation. 

I believe it's only designed to use by the proxy solution. The MultipartForm annotation is only targeted to parameter, method and field but not type. We have it on the api interface method.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 54 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HE5DSvaejhzyN95O:-KR6atn5_djSzsctHJLK:b-e0k82o) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L54)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> That sounds like a kludge. It might be better to switch completely to JSON, instead of switching to a mixture.
</details>
I think that switch (everything to json) might be better do it separately. 

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HN5L3GSShnPvoJEj:-KR6rbONGhejnQigzjaa:b-9k9lip) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L78)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Right, but can't we set the default to MediaType.APPLICATION_XML_TYPE and then override it for the few cases which use something else?
</details>
Looks like the default works (without giving it a specific type).

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 53 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR6N6yJ4h9aqEk8-2Z2:-KR6npFH2ZSYQY1IJWsD:b-896fix) ([raw file](https://github.com/zanata/zanata-client/blob/67153ea411b52c6c0e8c9b025d1a64621ae9e618/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L53)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Perhaps this should be ServerErrorException, as below?
</details>
Done.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 70 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR6NUSYTriP2KhIBCNz:-KR6nqCASkUb0oRz-fGD:b-896fix) ([raw file](https://github.com/zanata/zanata-client/blob/67153ea411b52c6c0e8c9b025d1a64621ae9e618/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L70)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> It might be better to create a custom exception here.
</details>
Done.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 78 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR6NZb9DbwJqnVgBcZh:-KR6pvW3exSrluVDRSbY:b-896fix) ([raw file](https://github.com/zanata/zanata-client/blob/67153ea411b52c6c0e8c9b025d1a64621ae9e618/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L78)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> It might be better to create a custom exception here, unless there's one in JAX-RS we could use.
</details>
Done.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java, line 91 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR6OC1-H41HQPbpfVRl:-KR6q4xmO621dXxVXu19:bj4sabs) ([raw file](https://github.com/zanata/zanata-client/blob/67153ea411b52c6c0e8c9b025d1a64621ae9e618/zanata-rest-client/src/main/java/org/zanata/rest/client/ResponseStatusFilter.java#L91)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> You could use `Response.Status.FOUND.getStatusCode()` or just use the numbers 301 and 302, but I think it would be best to pick constants or numbers, not one of each.
</details>
jaxrs api 2 has that status. I think it's missing in jersey (api 1?) class.  That's why the mixture. Now fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

seanf:
Reviewed 8 of 9 files at r3.
Review status: 37 of 38 files reviewed at latest revision, 3 unresolved discussions.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java, line 59 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2IgRFz_1vwusuayR3:-KR6wgiVymysMZrr1zuh:b-qf7wec) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/FileResourceClient.java#L59)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> I actually trace it down to its source code. The registered writer (in a map) need to match both the media type and the annotation.  We don't have a class like AnnotationLiteral in client so I came up with this idea to get hold of that specific annotation. 
> 
> I believe it's only designed to use by the proxy solution. The MultipartForm annotation is only targeted to parameter, method and field but not type. We have it on the api interface method.
</details>
Looks like `MultipartFormLiteral` is the way to go here.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 54 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HE5DSvaejhzyN95O:-KR6sgH6aZ2mSv6TanFX:b-bdgjpt) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L54)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> I think that switch (everything to json) might be better do it separately. 
</details>
Did this part of the API not exist when we used RESTEasy before? We didn't have this kludge before.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HN5L3GSShnPvoJEj:-KR6tPPG8YKjpltgX0YZ:b-yirrvi) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L78)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> Looks like the default works (without giving it a specific type).
</details>
Does that mean we are no longer sending an Accepts: header? If so, we can't rely on the server's behaviour.

Did you do anything with all the other instances?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

huangp:
Review status: 29 of 38 files reviewed at latest revision, 3 unresolved discussions.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 54 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HE5DSvaejhzyN95O:-KR6yQ4rNQM9wPxBnenh:bfm41r6) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L54)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Did this part of the API not exist when we used RESTEasy before? We didn't have this kludge before.
</details>
I don't remember. If it was it would be proxy based.  We do have those container type in dto. e.g. ProjectList. There is one called GlossaryResults which will do the trick if I have to use xml. But the name is misleading. 

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HN5L3GSShnPvoJEj:-KR75_pCLAvWnqi6yjd-:b-voi72f) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L78)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Does that mean we are no longer sending an Accepts: header? If so, we can't rely on the server's behaviour.
> 
> Did you do anything with all the other instances?
</details>
Ok reverted my changes. There is no way to set a default (that request method is only available after you constructed all your path in fluent api call chain)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

seanf:
Reviewed 8 of 8 files at r4.
Review status: 37 of 38 files reviewed at latest revision, 1 unresolved discussion.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HN5L3GSShnPvoJEj:-KR7KwnqC2v3PpVNSFpw:b-ch379o) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L78)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> Ok reverted my changes. There is no way to set a default (that request method is only available after you constructed all your path in fluent api call chain)
</details>
Shouldn't this say `request(MediaType.APPLICATION_XML_TYPE)` then?  We now have at least twelve instances of `request()` with no media type.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

huangp:
Review status: 37 of 38 files reviewed at latest revision, 1 unresolved discussion.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HN5L3GSShnPvoJEj:-KR7Ll5p1AcV94O99j8J:b-sk33r6) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L78)):*
<details><summary><i>Previously, seanf (Sean Flanigan) wrote…</i></summary>
> Shouldn't this say `request(MediaType.APPLICATION_XML_TYPE)` then?  We now have at least twelve instances of `request()` with no media type.
</details>
This one doesn't care about the response body. So the Accept header should not matter. Other places are either using streaming (the download file one) or deprecated (old synchronized SourceDocResource and TransDocResource). I did miss one though. Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

seanf:
Reviewed 1 of 1 files at r5.
Review status: 37 of 38 files reviewed at latest revision, 1 unresolved discussion.

---

*[zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java, line 78 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-client/126#-KR2HN5L3GSShnPvoJEj:-KR7NfY_czX987sFQTpJ:b-7ty478) ([raw file](https://github.com/zanata/zanata-client/blob/7370b34718f4d7bba1af0879c8f9e0c768a056ff/zanata-rest-client/src/main/java/org/zanata/rest/client/GlossaryClient.java#L78)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> This one doesn't care about the response body. So the Accept header should not matter. Other places are either using streaming (the download file one) or deprecated (old synchronized SourceDocResource and TransDocResource). I did miss one though. Fixed.
</details>
Okay, if you don't even look at the response, a missing `Accept` header is supposed to be equivalent to `Accept: */*`, so that should be fine for most of the PUT/POST/DELETE requests.

But in general, any methods which were using "Accept: xml" before (and they were back when we used generated proxies), we should keep sending "Accept: xml", at least until we switch to "Accept: json". A change of client library shouldn't be changing the communications between client and server.

SourceDocResource and TransDocResource still need an Accept header. And as far as I know they aren't deprecated. We still need a way of uploading and downloading generic content, whether we use XML or JSON for that.

The /file download methods are marked as `@Produces application/octet-stream` so it would be best to specify that (which is what we should get with generated proxies for FileResource). Media types still matter with streaming.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

seanf:
:lgtm:

---

Reviewed 4 of 4 files at r6.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126#-:-KRCnioD_dsLjQcpHx8r:bnfp4nl)*

seanf:
@huangp This has conflicts now.

seanf:
Reviewed 3 of 3 files at r7.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/126)*

PR #132, opened 2016-09-27
ZNTA-1271 - make static analysis happy

huangp:
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-client/132)


seanf:
I'm not sure we need to be changing client dependencies just to get past some rules in server/functional-test. The existing enforcer rules in server are mainly meant for zanata-war, not so much for functional-test, so we might need to relax the rules applied in functional-test.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-client/132#-:-KSeUNt96AbMjztiXNIf:bx9a0do)*

